PR #66, opened 2026-08-22
feat(git-read): read-only git_show/git_diff/git_log as child-safe tools

### LINKED ISSUE #61: Add read-only git tools (git_show/git_diff/git_log) to the child-safe tool surface
## Problem

Child agent types `reviewer` and `advisor` (tools: `read grep find ls fd rg`) have **no way to inspect git history**. In a real review session (OpenPI running on itself, review of PR #60's branch), both reviewer subagents explicitly reported (translated):

> "This session has no shell/exec tool and cannot run `git show`. The review is based on a full read of the HEAD working tree… all conclusions that depend on comparing against old code are marked as uncertain."

They could only review the *final state* of the tree, not the *diff*. Every claim about "behavior changed vs before" had to be cross-derived from a prose description the parent supplied, and the parent had to redo the diff review itself before merging.

## Why this matters

- Review is the highest-leverage use case for read-only child agents, and its primary artifact is a diff.
- The parent can't fix this by steering: the reviewer's tool restriction *correctly* excludes `bash`, so there is no way to grant git access without weakening the boundary.
- The current `CHILD_SAFE_PACKAGE_TOOL_NAMES` allowlist is exactly the right mechanism — it just lacks the capability that reviews need most.

## Proposal

Add a small read-only git surface as package tools, classified child-safe. Minimal set:

- `git_show` (commit contents, one revision)
- `git_diff` (two refs, or ref vs worktree)
- `git_log` (bounded output, e.g. `--oneline` + max count)

Constraints to respect (per the repo's own discipline):

- Read-only, fail-closed on anything that would write (validate the subcommand shape; don't shell out with interpolated strings).
- Must be registered in `CHILD_SAFE_PACKAGE_TOOL_NAMES` so the drift guard in `child-session.test.ts` stays authoritative.
- Follow the existing `file-search` pattern (structured schema, bounded output, spill-to-file on truncation).

tt-a1i:
Closes #61.

## Summary

Adds `extensions/git-read`: three read-only git tools (`git_show`, `git_diff`, `git_log`) classified as child-safe, so reviewer/advisor subagents can inspect commit history and diffs that their (correct) bash-free tool boundary previously excluded entirely.

The motivating failure was real: in the review round on #60, both reviewer subagents reported they could not run `git show` and had to review the final tree state instead of the diff, with the parent redoing the diff review before merging.

### Design

- **Follows the file-search pattern end to end**: structured schemas, bounded in-memory preview (2000 lines / 50KB), complete output persisted to a temp file up to the 10 MiB capture cap, process group terminated when the cap is exceeded, `sanitizeTerminalText` on delivered output.
- **Fail-closed argv construction** (`src/args.ts`, pure and unit-tested): revisions validated against a strict shape (sha / `HEAD~n` / name; leading `-` rejected so a revision can never smuggle a git flag), paths restricted to repo-relative segments (no `..`, no absolute), and every user value sits behind a `--` separator.
- **Child-safe by construction**: registered in `CHILD_SAFE_PACKAGE_TOOL_NAMES`; the source-scan drift guard in `child-session.test.ts` now enforces their classification. Only read subcommands are reachable — there is no code path that writes.
- **Built-in roles updated**: `explorer`/`reviewer`/`advisor`/`implementer` gain the git tools. Allowlists only ever narrow, so this widens nothing else.
- **Capability wiring**: new `gitRead` owner in `OPENPI_TOOL_SURFACE`; the `search` group now owns `fileSearch` + `gitRead`. The intent regex accepts 「用 git diff 比较」-style phrasing, and the advertised-phrase test pins it.
- **Replay safety extended** (found by e2e, not by inspection): `git_show`/`git_diff`/`git_log` are as side-effect-free as `fd`/`rg`, so they joined the replay-safe allowlist with source-path verification against `extensions/git-read/index.ts`. Without this, reviewer-type workflow agents silently stopped being replay-journaled — the e2e's child preflight failed on the missing tools before the fix, which is exactly the guard chain working as designed.

## Test plan

- [x] `args.test.ts` — argv purity: flag injection, path traversal, revision shapes, limit clamping
- [x] `index.test.ts` — real git repository: log/show/show+path/diff-worktree, error surfacing (bad revision, non-repo) as `GitCommandError` with stderr
- [x] Drift guard passes with the new tools classified
- [x] e2e (`execute.e2e.test.ts`) extended fake session exposes the git tools; preflight fail-closed was observed pre-fix
- [x] `replay-safety.test.ts` — git tools are replay-safe with the right source path
- [x] Full suite green: **772 node:test + 30 vitest**; `bun run check` green


tt-a1i:
合并前审查已完成并推送修复：

- 删除原提交中 62 个无关文件的 import 排序噪声，PR 只保留 Git 工具及必要集成。
- 修复能力门控：git-read 在 session_start 纳入 gitRead owner；explicit 普通会话保持 Bare Pi 工具面，加载 search 后 fd/rg/git_* 一次性出现。
- 禁用 --textconv / --ext-diff，防止仓库 .gitattributes + git config 让“只读”命令执行外部程序；增加真实仓库攻击回归。
- 启用工具自身 10s 超时与进程组清理；显式使用 10 MiB capture cap；Session Shutdown 清理保留的临时输出。
- to 无 from、staged 与 revisions 混用现在 fail-closed；双 ref 语义明确为 PR-style from...to。
- 修正 git_show(path) 合同：它限制 commit patch，而不是读取 blob。
- Ctrl+O 展开会显示最多 20 行真实结果；git_log 非 oneline 时不再把输出行误称为 commits。
- README/SETUP 与真实工具面同步。

最终基于最新 main 验证：796 Node tests + 30 Vitest 全绿；bun run check 通过（仅既存 file-search Effect warnings）。

PR #99, opened 2026-08-24
docs: reshape README around first-use success

tt-a1i:
## Summary
- reorganize the README around a quiet first viewport and three optional 3-minute success paths
- make explicit capability activation, observable completion, and parallel-write isolation visible before deep reference material
- consolidate repeated product explanations and fold low-frequency Workflow and development details without creating a second source of truth
- remove the rejected large Hero/runtime diagrams from the reading path and rely on portable GitHub Markdown

## Validation
- `git diff --check`
- Impeccable README detector: no findings
- internal navigation and referenced local files checked

## Notes
- documentation-only change; runtime and package behavior are unchanged
- local `.impeccable/` review artifacts are intentionally excluded

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <p align="center">
-  <img src="assets/openpi-package.png" alt="OpenPI logo" width="240" />
+  <img src="assets/openpi-package.png" alt="OpenPI logo" width="180" />
 </p>
 
 <h1 align="center">OpenPI</h1>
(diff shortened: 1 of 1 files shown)


tt-a1i:
Thanks for the thoughtful README restructuring. This branch now conflicts with the current main and would reintroduce outdated documentation, so I'm closing it rather than resolving the conflict in place. The focused onboarding improvements can be reapplied in a fresh change based on the latest main.

PR #147, opened 2026-08-25
fix(workflows): make agent startup abortable

### LINKED ISSUE #116: workflows: agent startup (session creation + extension binding) is not abortable
## Description

Verified by read-only audit (`seal/gpt-5.6-sol` subagent, verdict: CONFIRMED — low severity, lifecycle hardening).

## Mechanism

- `runAgent` awaits `sessionFactory ?? createAgentSession` (`runner.ts:532-544`) then `bindChildSessionExtensions` (`runner.ts:545`), which awaits `session.bindExtensions({ mode: "print" })` (`extensions/shared/child-session.ts:215-219`) — binding serially awaits `session_start` and `resources_discover` extension handlers (`pi-coding-agent/dist/core/agent-session.js:1741-1762`).
- The abort listener is installed only afterward (`runner.ts:652-684`).
- Concrete handlers can be slow: `extensions/file-search/index.ts:144-174` awaits binary initialization, which can probe subprocesses, download over HTTPS, and invoke `tar` (`extensions/file-search/src/binaries.ts:182-220,327-407`) with internal timeouts.

## Impact (bounded)

Cancellation cannot interrupt either startup await; a genuinely hung handler can outlive the controller's 8-second settlement deadline (`controller.ts:165-187`). However, once startup completes, the late `signal.aborted` check (`runner.ts:680-683`) aborts/disposes the session and `if (!aborted)` skips `prompt()` (`runner.ts:689-704`) — the agent does **not** continue running after cancellation.

Net impact: delayed/unbounded startup cleanup and possible settlement timeout, not post-cancellation prompting. Also, a startup rejection caused by cancellation is caught as `aborted: false`, slightly misclassifying cancellation (still settles).

## Suggested fix

Install/observe the signal before session creation and race session creation/binding against an abort promise, disposing any session created after cancellation. Return an aborted outcome from startup.

Found by automated workflow audit (run `wf_56028a5b018b`).

CuSO41108:
## Summary

Closes #116.

Make workflow child-agent startup responsive to cancellation during both session creation and extension binding.

When cancellation wins the startup race, `runAgent()` now returns an explicit aborted outcome without waiting indefinitely. Any session created after cancellation is still owned and cleaned up, and late startup rejections are observed.

## Changes

- Install the abort listener before creating the child session.
- Race session creation and extension binding against cancellation.
- Skip session creation when the signal is already aborted.
- Abort and dispose sessions that are created after cancellation.
- Reuse the existing bounded, idempotent child-session cleanup path.
- Prevent prompting after startup cancellation.
- Classify startup cancellation as `aborted: true`.
- Observe late factory and binding rejections to avoid unhandled rejections.

`bindExtensions()` does not currently accept an `AbortSignal`, so its internal initialization may still finish in the background. This change stops the workflow from waiting on it indefinitely and safely handles its eventual result.

## Tests

Added deterministic regression coverage for:

- cancellation before startup;
- cancellation while session creation is pending;
- cleanup of a session created after cancellation;
- a late session-factory rejection;
- cancellation while extension binding is pending;
- preventing prompts and duplicate disposal after cancellation.

Validation:

- `extensions/workflows/runner.test.ts`: 21/21 passed
- `bun run lint`: passed
- `bun run typecheck`: passed

Local full-suite validation is affected by existing Windows-specific failures involving process handling, path separators, file modes, and symlink permissions. The runner test suite and all new regression cases pass.


tt-a1i (approved):
Thanks for the focused lifecycle hardening. I reviewed the startup cancellation races across pre-aborted startup, pending session creation, late factory settlement, extension binding, prompting, and bounded idempotent cleanup. The branch also integrates cleanly with current main; bun run check, the 21 runner tests, and the full suite (874 node tests plus 30 file-search tests) pass. No blocking issues from my side.

PR #149, opened 2026-08-25
fix(background-terminals): harden output and process lifecycle

somewan820:
## Summary

- sanitize status, single-result, and batched model-facing terminal output before tail truncation without mutating retained or spill evidence
- enforce a 512 MiB per-session spill budget on top of the per-stream limit, fail closed when complete logs become unavailable, and remove pruned spill files
- await bounded Windows `taskkill` results, serialize graceful and force phases, preserve natural-exit races, and avoid claiming an unconfirmed process-tree kill
- add focused budget and Windows lifecycle coverage, a `windows-latest` CI job, and align the implementation guide with the runtime contract

## Validation

- `bun run check`
- `bun run test` (888 Node tests passed, 1 Windows-only test skipped on macOS; 30 Vitest tests passed)
- workflow YAML parsed locally; the new Windows runner provides the real `taskkill /T` process-tree validation

Closes #72

PR #148, opened 2026-08-25
refactor(setup): remove pi-intercom ownership

sunsunsun-java:
## Summary

- remove pi-intercom detection, recommendation, installation, configuration writes, and setup status from OpenPI
- document pi-intercom as an independently managed top-level Pi Session package while preserving child-session isolation
- use Pi's canonical package-source matching and fail closed before local package factories execute when pi-intercom identity cannot be verified
- preserve ordinary package provenance, lifecycle hooks, resource filtering, and collision handling

## Testing

- bun run check
- bun run test (Node: 864 passed; Vitest: 30 passed)

Closes #128

PR #151, opened 2026-08-25
refactor(test): move suites into top-level tests directory

### LINKED ISSUE #150: refactor(test): 将测试迁移到顶层 tests/ 并建立递归发现合同
## 问题

当前 OpenPI 的测试文件与生产 TypeScript 文件共置在各个 `extensions/<name>/` 目录中。仓库目前共有 238 个 TypeScript 文件，其中 107 个是 `.test.ts` / `.spec.ts`，在测试密度较高的模块中，测试已经明显影响源码目录的浏览与职责识别。

这不只是目录观感问题，还暴露了几个工程合同缺口：

1. 当前测试命令使用：

   ```text
   node --test --experimental-strip-types extensions/*/*.test.ts
   ```

   它只发现 extension 根目录下一层测试；未来放入 `src/` 或其他子目录的测试会被静默漏跑。该问题已在 #73 中确认。

2. 测试专用 stub、fixture 或 compatibility helper 缺少明确归属，容易留在 production module 中并进入 npm artifact。#130 已记录相关实例。

3. 根 `tsconfig.json` 当前只 include `extensions/**/*.ts`。迁移测试目录后，需要明确保证所有测试仍进入 typecheck。

4. 目录级配置正在由 #131 收敛，测试目录也应有单一、清晰的根级工程合同，而不是继续增加 extension-local 配置。

## 目标

建立顶层 `tests/` 作为测试代码的唯一归属，并按生产模块镜像组织：

```text
tests/
  extensions/
    workflows/
    subagents/
    shared/
    ...
  support/
  fixtures/
  e2e/
```

建议保持测试文件与真实 owner 的对应关系，例如：

```text
extensions/workflows/runner.ts
tests/extensions/workflows/runner.test.ts
```

测试 support、stub 与 fixture 应留在 `tests/support/` 或 `tests/fixtures/`，不为测试向生产模块增加 export。

## 实施边界

- 迁移现有 `.test.ts` 与 `.spec.ts` 到顶层 `tests/`，保留 extension/module 映射。
- 建立递归、确定性的测试发现，确保新增嵌套测试会进入本地与 CI。
- 保留现有 `node:test` 与 Vitest 分工；本 issue 不迁移测试框架。
- 更新根 TypeScript 配置，使 `extensions/**/*.ts` 与 `tests/**/*.ts` 都进入 typecheck。
- 将仅供测试使用的 stub、fixture、helper 移入测试目录；与 #130 重叠的改动应避免重复实现。
- 验证 npm package 内容，确保 `tests/` 及 test-only support 不进入发布产物。
- 保持用户可见行为、模型可见上下文和运行时/持久化行为测试的边界清晰；目录迁移不应合并这些不同层级的断言。

## 建议实施顺序

1. 先修复递归测试发现，并加入一个嵌套测试确实被执行的守卫。
2. 再机械迁移现有测试及其相对 import。
3. 最后移动 test-only support，并验证 npm artifact。

这样可以避免出现“文件已经搬迁，但测试命令没有执行它们”的假绿色状态。

## 不包含

- 不改变生产运行时行为。
- 不统一或替换 `node:test` 与 Vitest。
- 不为了方便测试而扩大 production API。
- 不在本 issue 中处理 #73 的 peer dependency、跨 extension import、worktree cleanup 或 `as any` 等其他事项。
- 不以大规模源码目录重构或拆分 extension 为目标。

## 验收标准

- [ ] 所有测试位于顶层 `tests/`，并能清楚映射到对应 extension/module。
- [ ] 任意 `tests/**/*.test.ts` 都会进入本地与 CI 的 Node test suite。
- [ ] Vitest 测试也通过稳定、明确的模式被发现。
- [ ] 存在防止嵌套测试静默漏跑的自动化守卫。
- [ ] 根 typecheck 覆盖全部生产源码和测试源码。
- [ ] 测试专用 stub、fixture 与 helper 不再混入 production module。
- [ ] `npm pack` 产物不包含 `tests/` 或其他 test-only support。
- [ ] 迁移前后的测试数量与测试语义保持一致；如有增删，必须逐项说明。
- [ ] `bun run check` 与 `bun run test` 通过。

## 关联

- #73：测试发现盲区
- #130：test-only compatibility facade 与 stub
- #131：根级 TypeScript 配置收敛

somewan820:
## Summary

- move all existing test files into a top-level tests/ mirror
- add recursive, deterministic discovery for Node test and Vitest suites
- include tests in root typecheck and Biome coverage
- move the test-only Subagent stub into tests/support/
- document the new test ownership and discovery contract

## Validation

- `bun run check`
- `bun run test` — Node 875/875, Vitest 30/30
- `npm pack --ignore-scripts --dry-run` — 149 package files; no `tests/`, `test-support`, or `subagents-stub` entries
- `git diff --check`

TypeScript emits the repository's existing Effect language-service warnings only.

Closes #150
Related to #73 and #130.


tt-a1i (dismissed):
整体迁移做得很完整：顶层 tests mirror、递归发现、根 typecheck、Biome 和 package exclusion 都对齐了；当前 Linux CI、本地 881 个 Node tests、30 个 Vitest tests 以及 npm pack 检查也都通过。

有一个跨平台阻断问题需要修复：`scripts/run-tests.mjs` 通过 `spawnSync()` 直接执行 `node_modules/.bin/vitest`。Windows 安装下这个入口通常依赖 `.cmd` shim，而 Node 的无 shell child process 不能直接启动 `.cmd` 文件，因此 `bun run test` 会在 Node suite 完成后无法启动 Vitest。当前 #151 只有 Linux Node 22/24 CI，没有覆盖该路径。

建议改为通过 Vitest 的跨平台 JavaScript entrypoint 启动，或使用 Bun 支持的本地 executable 调用方式，并增加一个 Windows runner/验证。Node 官方说明：https://nodejs.org/api/child_process.html#spawning-bat-and-cmd-files-on-windows

另请协调 #149：它新增的 Windows job 仍引用 `extensions/background-terminals/manager.test.ts`，而本 PR 会将该文件移到 `tests/extensions/background-terminals/manager.test.ts`。两个 PR 组合前需要同步更新路径，避免各自单独 CI 通过、合并后 Windows job 找不到测试文件。

除此之外没有发现其他阻断问题。

somewan820:
已处理本次 review 与主线冲突，说明如下：

- 已将最新 `origin/main` 合入本 PR，当前分支包含主线的 workflow sandbox IPC 参数 fail-closed 修复及最新 CI/CODEOWNERS 变更。
- Windows runner 问题已在提交 `1829655` 修复：`scripts/run-tests.mjs` 通过 `process.execPath` 启动 `node_modules/vitest/vitest.mjs`，不再直接执行 Windows `.cmd` shim。
- #149 的 Windows job 已采用兼容路径：优先 `tests/extensions/background-terminals/manager.test.ts`，同时兼容旧路径，因此与本 PR 合并后的测试目录迁移不会失效。
- 验证结果：`bun run check` 通过；`bun run test` 通过（887 Node tests、30 Vitest tests）。
- 本次冲突解决后的合并提交：`b6b74bd`。

请重新检查并更新 review 状态。

somewan820:
补充同步：主线刚合入 #148（a023dac），该提交移除了 OpenPI 对 pi-intercom 的所有权。已再次将最新 origin/main 合入本 PR，按主线决策删除了迁移后遗留的 tests/extensions/setup/intercom.test.ts；其余测试目录迁移与 Windows runner 修复保持不变。最新同步提交：035c66d。最新验证：bun run check 通过；bun run test 通过（871 Node tests、30 Vitest tests）。

tt-a1i (dismissed):
Thanks for the thorough test-layout cleanup—the implementation itself looks sound, and I found no code-level blocker. One evidence issue needs correction before approval: the PR currently reports the Node suite as 881/881, but ✔ interaction tools follow overlapping Plan and Setup modes (2.779292ms)
✔ formats selected, noted, and custom answers (1.16825ms)
✔ preview truncation preserves Unicode characters and strips controls (11.352291ms)
✔ prompt requires genuine ambiguity, recommendations, and no continue questions (0.105ms)
✔ dismissal does not imply an answer (0.183667ms)
✔ model-facing answers strip terminal controls and structural newlines (0.132042ms)
✔ answer drafts have a UTF-8 byte bound (1.156584ms)
✔ review summaries are single-line, terminal-safe, and bounded (0.107833ms)
✔ option labels with the same terminal-safe display identity are rejected before UI (0.731083ms)
✔ visibly distinct option labels remain valid (2.03375ms)
✔ questionnaire rendering strips controls and wraps long CJK text without dropping it (5.730333ms)
✔ answers stay draft until review, where any question can be revised (0.535459ms)
✔ free-form text survives leaving and reopening its draft editor (1.544167ms)
✔ active TUI drafts strip terminal and bidi spoofing controls (1.252542ms)
✔ compact paste content survives Escape and reopening the draft (0.574917ms)
✔ a blank free-form draft explicitly requests a rephrased question (0.510583ms)
✔ an oversized compact paste is rejected without clearing the existing draft (1.422708ms)
✔ split oversized bracketed paste is discarded before Editor accumulates it (0.696708ms)
✔ notes survive leaving and reopening their draft editor (0.50175ms)
✔ dismissing the review returns no draft answers (0.119292ms)
✔ interaction tools are sequential and expose bounded usage guidance (0.096792ms)
✔ RPC dialogs preserve drafts, allow revision, and require explicit review (0.323375ms)
✔ RPC dialogs show an option preview before accepting that choice (0.1225ms)
✔ human handoff reviews status and tells the agent to verify completion (0.254459ms)
✔ session start keeps only the background entry tool active (4.860833ms)
✔ a successful start exposes lifecycle tools but a rejected start does not (1025.007208ms)
✔ happy path: stdout and stderr captured separately, settles done, hook fires once unconsumed (1095.289625ms)
✔ non-zero exit settles as failed with the exit code (1090.497208ms)
✔ optional deadline terminates the process tree and settles timed_out (1060.152458ms)
✔ kill settles a never-exiting process as killed and resolves after settle; repeat kill is a no-op (1012.260166ms)
✔ a SIGTERM-resistant child is escalated to SIGKILL within the teardown bound (3108.572041ms)
✔ concurrent overlapping multi-id kills observe each settlement exactly once (1012.0115ms)
✔ kill terminates the whole process tree (grandchildren die) (1141.97525ms)
✔ a shell exit with inherited pipes open settles naturally and reaps descendants (1068.900333ms)
✔ kill preserves a natural exit observed before the signal point (1035.247583ms)
✔ concurrency cap rejects an extra start; a failed spawn releases its slot (1169.512458ms)
✔ a settle during an in-flight kill reports consumed: true (1007.892833ms)
✔ UI requestKill settles as killed and is NOT consumed (1013.704709ms)
✔ runtime.dispose kills running processes; no settle hook fires after dispose (1019.237458ms)
✔ pruning drops the oldest settled entries past MAX_TRACKED, never running ones (1436.73025ms)
✔ runtime disposal removes the private spill directory (1071.063417ms)
✔ an unknown command settles failed with the platform shell's non-zero exit (1024.299334ms)
✔ a process 'error' event settles failed with errorText and no bogus exit code (2.359542ms)
✔ the spill file holds the complete capture when the settle hook fires, beyond the in-memory cap (1064.343417ms)
✔ aborting the kill wait does not cancel the termination (3116.507417ms)
✔ status returns the snapshot and rejects unknown ids with the known list (1007.249167ms)
✔ push/view roundtrip preserves text and counts bytes (0.798125ms)
✔ head chunks are evicted past the cap and accounted as truncated (0.076375ms)
✔ a single chunk larger than the cap is trimmed to its tail — retention stays bounded (0.112959ms)
✔ an oversized chunk evicts everything retained before it (0.065625ms)
✔ an oversized chunk cut lands on a UTF-8 code point boundary (0.081791ms)
✔ spill receives the complete oversized chunk before trimming (2.271125ms)
✔ push reports spill backpressure while retaining the chunk (0.08925ms)
✔ byte accounting uses UTF-8 byte length, not string length (0.051917ms)
✔ multibyte chunks are never split by eviction (0.069416ms)
✔ spill callback receives every chunk in order, even after eviction (0.103959ms)
✔ view text is cached between pushes and version increments per push (0.066666ms)
✔ start descriptions identify the platform-specific shell contract (0.700666ms)
✔ kill report distinguishes killed / raced natural exit / already settled (0.106208ms)
✔ status result marks head-truncated output with a pointer at the full log (0.266542ms)
✔ completion message reports kill, timeout, or exit and omits empty stderr (0.146334ms)
✔ batched completion preserves identities and globally bounds log tails (5.108042ms)
✔ completion output is a shorter tail than the detailed status view (0.142791ms)
✔ dashboard selection follows its terminal id and falls back by row (1.320375ms)
✔ sanitizeText strips ANSI, tabs, and control characters (0.297958ms)
✔ output line cache reuses a version/width key and invalidates either dimension (2.015875ms)
✔ buildOutputLines wraps long lines and keeps only the final CR segment (0.174ms)
✔ buildOutputLines drops one trailing empty line from a trailing newline (0.078417ms)
✔ start capacity counts running, retractable results, and concurrent reservations (1.966708ms)
✔ a result consumed by a kill/status is not delivered (0.755584ms)
✔ defer reports pending cardinality without evicting identities (0.121125ms)
✔ unconsumed results are delivered once in settlement order (0.060791ms)
✔ bounded drains keep overflow retractable and preserve retry order (0.119667ms)
✔ re-deferring the same id replaces rather than duplicates (0.055791ms)
✔ a drained result can be retained for retry after delivery fails (0.047292ms)
✔ idle result batching uses one fixed bounded window (0.112333ms)
✔ a full pending batch flushes immediately with the current wake policy (0.090916ms)
✔ a busy race retains the batch for a later quiet flush (0.113958ms)
✔ clearing an idle result batch prevents a post-shutdown flush (0.101209ms)
✔ only a result the model is waiting for costs it a turn (0.059958ms)
✔ background terminal preview keeps only each stream tail (1.864417ms)
✔ background terminal preview reports upstream truncation without showing its long notice (0.141708ms)
✔ compact renderer stays bounded, shows the tail, and advertises expansion (2.338916ms)
✔ batched terminal rendering preserves identity without inventing a singular terminal (0.239916ms)
✔ background terminal preview omits empty streams (0.069708ms)
✔ matches within a single chunk and reports the containing line (1.55975ms)
✔ matches retained output when armed after a fast process spoke (0.732375ms)
✔ refuses to arm after eviction makes future control state unknowable (0.206958ms)
✔ skips retained streams whose missing head makes control state unknowable (0.064583ms)
✔ matches a pattern that straddles two chunks (0.084166ms)
✔ is one-shot: never fires twice even without disarming (0.072625ms)
✔ supports an alternation covering failure signatures (0.077083ms)
✔ bounds retained carry so a firehose cannot grow memory (0.71925ms)
✔ matches literal signatures without exposing the regex hot path (0.076167ms)
✔ rejects unsafe or ambiguous signature lists (0.126542ms)
✔ keeps stdout and stderr carries separate (0.062375ms)
✔ strips ANSI, OSC, and C1 control bytes from the reported line (0.0865ms)
✔ never matches signatures hidden inside terminal control payloads (0.049125ms)
✔ keeps split long control payloads hidden across the carry bound (0.130083ms)
✔ bounds the reported line when the stream has no newline (0.066125ms)
✔ ordinary sessions add no resident OpenPI model tool (1.884375ms)
✔ adaptive discovery keeps only the capability gateway resident (2.052667ms)
✔ setup changes capability discovery immediately without unloading groups (0.427541ms)
✔ the gateway does not add an ineffective resident adoption guideline (0.095416ms)
✔ the gateway exposes a provider-portable string enum for capability groups (0.148375ms)
✔ the gateway stays within its compact provider-surface budget (0.084875ms)
✔ an explicit OpenPI request exposes the capability gateway (0.144666ms)
✔ an explicit subagent request loads delegation directly (0.229541ms)
✔ reserved capability words load their groups without action verbs (0.156416ms)
✔ common Chinese and multi-agent delegation requests are explicit intent (3.589208ms)
✔ the README quick-start example and its advertised phrases load capabilities (1.316875ms)
✔ the gateway returns progressive skill guidance for a loaded group (0.273375ms)
✔ conditional or negated capability boilerplate does not widen the tool surface (1.83475ms)
✔ direct delegation imperatives and questions still load delegation (0.334416ms)
✔ capability loads are monotonic and activate only the requested group (0.205459ms)
✔ uses Claude-style lavender with a contrast-safe light variant (0.787334ms)
✔ highlights capability names only when shared intent authorizes them (5.618ms)
✔ supports English plurals and Chinese capability names (0.082791ms)
✔ render feedback never mutates the submitted editor text (0.058833ms)
✔ ghost suggestions stay dim until the user accepts them into the editor (2.184667ms)
✔ context_pivot is visible only when the next run starts above its threshold (2.714542ms)
✔ estimates context from exact tokens or percentage (0.079583ms)
✔ builds a clean pivot summary without notebook or handoff coupling (0.129209ms)
✔ cron fires only while idle and retries a failed one-shot delivery (2.910417ms)
✔ cron batches every job due in one tick into one triggered turn (0.296625ms)
✔ a recurring cron interval starts at successful delivery time (0.102584ms)
✔ a failed cron batch leaves every due job pending for one retry (0.112917ms)
✔ cron shutdown clears jobs and stops future delivery (0.098208ms)
✔ parses duration units and rejects nonsense (0.816ms)
✔ parses recurring, one-shot, and management commands (0.922584ms)
✔ rejects an interval finer than the poll can honor (0.127709ms)
✔ selects only jobs that are due (0.080625ms)
✔ a recurring job reschedules from now, so a busy gap cannot burst (0.071334ms)
✔ a one-shot job drops out after firing (0.042375ms)
✔ a failed delivery remains due instead of losing the prompt (0.079042ms)
✔ formats intervals back to their friendliest unit (0.05375ms)
✔ overrides all seven activity renderers without changing model-facing definitions (397.248167ms)
✔ real ToolExecutionComponent toggles between one activity row and native evidence (2.24775ms)
✔ all activity tools render one semantic success row (4.45175ms)
✔ completed targets are muted without weakening pending or failed targets (7.214833ms)
✔ wrapping changes only renderer slots and shell ownership (0.441083ms)
✔ all activity tools render pending and failure as one explicit row (3.557ms)
✔ long activity rows stay one line and fit narrow terminals (0.245708ms)
✔ successful truncation ellipses use the target's muted foreground (0.402625ms)
✔ expanded mode delegates call and result to Pi native renderers (7.000791ms)
✔ repository text delegates to the canonical terminal sanitizer (1.609125ms)
✔ automatic refresh stays local until /pr explicitly requests GitHub data (543.725542ms)
✔ captures output and tolerates command failures (165.913542ms)
✔ renders platform failures without making callers handle them (12.886709ms)
✔ reports command timeouts as failures (24.489833ms)
✔ an explicit refresh waits for an active background refresh (2.659125ms)
✔ buildShowArgs validates revisions and paths (2.054541ms)
✔ buildDiffArgs composes only read-only diff forms (0.157166ms)
✔ buildLogArgs clamps the limit and validates inputs (0.234292ms)
✔ runGit returns stdout for a successful command (32.47125ms)
✔ runGit surfaces git failures as GitCommandError with stderr (23.248167ms)
✔ git show output includes commit metadata and patch (32.6815ms)
✔ git show with path limits the commit patch to that path (25.0845ms)
✔ structured git diff and show never invoke diff.external (221.990125ms)
✔ structured git diff and show never invoke textconv commands (394.181958ms)
✔ runGit applies its own bounded timeout (32.014875ms)
✔ structured git diff preserves worktree, staged, revision, and path forms (147.917667ms)
✔ outside a repository the command fails with a clear error (44.970791ms)
✔ expanded git results show a bounded preview and retained output path (0.686667ms)
✔ assistant accounting matches Codex non-cached input plus output and falls back for legacy usage (0.838125ms)
✔ v2 active goals survive restore while v1 active goals migrate to a durably paused v2 state (9.09925ms)
✔ kickoff queues one Codex-style continuation without consuming usage (1.312542ms)
✔ a goal created during a busy run accounts post-create work and continues after the absorbed steer (0.486084ms)
✔ a settled active goal accounts tokens and time then dispatches exactly one continuation (0.180458ms)
✔ user interruption pauses, provider errors block, usage errors limit, and overflow retry aborts do neither (0.515416ms)
✔ crossing a token budget marks the goal limited and sends only a wrap-up turn (0.737375ms)
✔ editing uses a plain continuation when idle and objective-updated steering while busy (0.31225ms)
✔ busy editing a complete goal tracks only post-edit work and keeps the loop running (0.213292ms)
✔ completion acknowledgement follows a tagged user message and persists after it (3.023042ms)
✔ handled explicit input does not acknowledge a later extension user message (0.122958ms)
✔ an extension-reinjected tagged input cannot acknowledge after its correlation settles (0.106125ms)
✔ fork deferral survives restore until explicit input releases it (0.12275ms)
✔ model completion flushes current assistant usage into its tool result without double accounting (0.167334ms)
✔ blocked reports require the same blocker on three consecutive distinct goal turns (0.37125ms)
✔ a different or unreported blocker resets the blocked audit (0.240625ms)
✔ the durable blocked audit survives reload and starts fresh after resume (0.249125ms)
✔ model blocking cannot demote a budget-limited goal (0.05675ms)
✔ resume accepts paused, blocked, and usage-limited but not budget-limited (0.262ms)
✔ failed continuation dispatch durably pauses the active goal (0.089084ms)
✔ a failed continuation dispatch does not increment the continuation count (0.88ms)
✔ resume resets the continuation count so the emergency breaker can recover (0.169583ms)
✔ goal command matches the current Codex surface (1.294333ms)
✔ bare goal summary uses Codex fields, compact usage, and status-specific command hints (0.598625ms)
✔ tool responses expose only Codex-shaped goal fields and remaining budget (0.135ms)
✔ goal lifecycle tools appear only after goal state exists (4.824625ms)
✔ update_goal requires a blocker description and returns structured rejected-audit feedback (0.856583ms)
✔ model tools create, read, complete, and replace goals with Codex semantics (0.528667ms)
✔ slash command creates directly, confirms unfinished replacement, edits, pauses, resumes, and clears (0.694083ms)
✔ the next explicit user message durably hides an achieved footer without clearing goal history (0.585834ms)
✔ restored active goals auto-continue, stopped goals prompt, and forks defer until explicit input (1.634333ms)
✔ goal prompts match Codex lifecycle guidance and XML-escape user objectives (0.785583ms)
✔ creates Codex-shaped v2 goals with a 4000-character objective and positive optional budget (1.724083ms)
✔ goal input and restored snapshots strip terminal and bidi spoofing controls (0.243542ms)
✔ v1 migration pauses active work once, folds distinct success criteria, and maps terminal statuses (8.079584ms)
✔ restore remains branch-local, ranks by revision then position, and locks after later malformed history (0.295167ms)
✔ progress accounting, deferral, continuation count, and token budget are durable transitions (0.212125ms)
✔ blocked audit validates durable current-turn state and ignores duplicate reports (0.7795ms)
✔ completion acknowledgement is durable, idempotent, and complete-only (0.135625ms)
✔ editing preserves running and stopped statuses but reactivates complete or budget-limited goals (0.212917ms)
✔ token and elapsed-time formatting matches current Codex Goal UI (1.286125ms)
✔ footer never exposes the objective or legacy turn counters (0.607959ms)
✔ goal tool result paints every row through the supplied status background (1.695458ms)
✔ goal rendering sanitizes replayed objective, reason, status, and messages (0.751ms)
✔ continuation rows are one-line labels and never reveal prompt bodies (0.088417ms)
✔ the investigation commands a plan is actually built on are allowed (1.884042ms)
✔ raw git commands that render diffs are refused (0.844791ms)
✔ git log stays available only when it does not render a diff (0.212834ms)
✔ a value-taking global flag cannot donate the subcommand to itself (0.150917ms)
✔ gh cannot have its read verb swallowed by a preceding flag (0.092ms)
✔ git flags that run a program or write a file are refused (0.082125ms)
✔ an unrecognized flag is refused rather than assumed harmless (0.081ms)
✔ quoted arguments and read-only search tools are allowed (0.088625ms)
✔ shell composition, expansion, and unsafe input syntax are refused (0.128292ms)
✔ search and inspection flags remain effect allowlisted (0.157083ms)
✔ unknown programs explain the available read-only commands (0.061541ms)
✔ write subcommands stay refused now that they cannot be reached sideways (0.100458ms)
✔ pathspecs after -- are not mistaken for flags (0.1635ms)
✔ tilde is refused where a shell expands it, not inside a revision (0.074042ms)
✔ a refusal explains itself, since the model has to react to it (0.078417ms)
✔ a missing or non-string command is refused rather than assumed empty (0.0705ms)
✔ plan mode allows only explicit observational tools (0.821584ms)
✔ entering Plan Mode loads the structured search capability (1.327791ms)
✔ plan mode fail-closes known mutations and future unknown tools (0.079459ms)
✔ bash is judged per command, not blocked as a whole tool (0.369917ms)
✔ plan_ready fails preflight unless it is the only tool call (0.101666ms)
✔ Plan Ready keeps the write gate closed until the user prepares an editable implementation prompt (1.111125ms)
✔ bare /plan offers to turn planning mode off (0.208834ms)
✔ bare /plan offers to turn ready mode off (0.205209ms)
✔ /plan off clears plan mode (0.156875ms)
✔ fresh implementation links a new session and prefills without submitting (0.287791ms)
✔ spawning is allowed while planning, resuming an existing child is not (0.100709ms)
✔ planning children get investigation tools and never gain any (0.310625ms)
✔ plan mode rejects type declarations it would silently narrow (0.068792ms)
✔ persisted Plan Mode state restores branch-locally and fails closed (0.168584ms)
✔ session reload and tree navigation restore the branch-local write gate (0.356042ms)
✔ the stance is broadcast on every change, including shutdown (0.798792ms)
✔ post-edit runs only in interactive TUI mode (0.960083ms)
✔ post-edit serially drains changed turns that settle during a run (1.802625ms)
✔ post-edit sanitizes failure notifications (0.841208ms)
✔ post-edit independently bounds the command and output in failure notifications (2.111916ms)
✔ post-edit bounds execution errors before notifying (0.989625ms)
✔ post-edit aborts an in-flight command on session shutdown (0.199084ms)
✔ large session lists bound Git subprocesses and reuse cached stats (12.881875ms)
✔ cancelling a stale list stops queued work and prevents stale cache updates (2.371375ms)
✔ session labels prefer names and limits reject invalid input (0.973792ms)
✔ session search matches normalized metadata (0.116417ms)
✔ persisted session labels and preview content are terminal-safe (1.558583ms)
✔ setup broadcasts its episode demand for interaction tools (3.218709ms)
✔ registers the canonical setup command, legacy alias, and one constrained tool (0.426667ms)
✔ post-edit stays off or preserved unless the setup request changes it (147.091667ms)
✔ legacy ui_footer_items stays an input adapter and persists only footerLines (31.10675ms)
✔ session_start hides configure_my_pi_setup after registration refresh (0.756833ms)
✔ openpi-setup and my-pi-setup expose the tool then inject the setup message (1.306375ms)
✔ successful configure_my_pi_setup hides the tool (0.634583ms)
✔ successful setup result closes the episode and names the only re-entry (0.158ms)
✔ keep-without-apply hides after the setup agent run settles (0.478541ms)
✔ follow-up while busy does not tear down on the prior turn settle (0.416875ms)
✔ overlapping openpi-setup follow-up survives the prior apply (0.437834ms)
✔ a second openpi-setup after hide re-exposes via get+add (0.426708ms)
✔ session_start after host re-includes tools hides configure tool (0.40825ms)
✔ setActiveTools never snapshot-restores unrelated tools away (0.387583ms)
✔ failed configure_my_pi_setup stays visible until the setup run settles (0.289042ms)
✔ offers optional Intercom only for an idle argument-free TUI setup (0.049333ms)
✔ the role-model schema exposes every built-in role as an optional property (0.033542ms)
✔ builds a model-guided first-run setup prompt with impacts (0.274ms)
✔ partially assigns and clears validated subagent role models (0.236208ms)
✔ builds a focused review prompt when configuration already exists (0.050166ms)
✔ recognizes only the pi-intercom npm package identity (0.8665ms)
✔ new installs receive private safe defaults before package activation (434.697708ms)
✔ existing preferences and unknown fields are preserved (321.698209ms)
✔ an existing config missing safe fields is preserved and blocks installation (189.506ms)
✔ malformed or unsafe typed config blocks installation without overwriting (769.360041ms)
✔ a symlinked intercom directory is rejected before writing outside agentDir (34.256667ms)
✔ a symlinked config is rejected instead of overwriting its target (309.710125ms)
✔ a directory swap during package download aborts before config or activation (134.873875ms)
✔ the filesystem helper rejects a swapped cwd inode before writing (135.125875ms)
✔ a manual config created during package download is preserved exactly (275.459458ms)
✔ concurrent OpenPI installers fail closed on the package config lock (473.23675ms)
✔ failed package download writes no config (278.608833ms)
✔ failed package-setting persistence retains newly created safe defaults (556.552375ms)
✔ persistence failure preserves a concurrently replaced safe config (439.681625ms)
✔ inspects configured package version and effective safety defaults (81.854875ms)
✔ formats absent, installed, and invalid integration states (0.194333ms)
✔ Pi host packages stay peers while local checks keep development copies (1.129958ms)
✔ the manifest enforces the documented Node floor (0.073042ms)
✔ pi-intercom stays an explicit opt-in instead of a bundled dependency (0.049875ms)
✔ the public OpenPI package has complete gallery and registry metadata (0.680916ms)
✔ running work always reports; settled work only until the next request (2.130125ms)
✔ acknowledged settled work leaves no status line (0.094167ms)
✔ a settle without a timestamp is treated as unread (0.056917ms)
✔ status text names its own view command (0.099ms)
✔ Direct and Workflow children use one read-only full-terminal page (15.177042ms)
✔ a read-only Workflow child returns left without stealing a parent session (0.12725ms)
✔ nested below-editor strips are reachable with repeated Down (1.882625ms)
✔ a newly available outer strip can take focus from a nested strip (0.102083ms)
✔ Down remains consumed when a focused strip has no nested peer (0.074083ms)
✔ classifies explicit capability intent across English, Chinese, and mixed phrases (5.90225ms)
✔ reserved English capability words authorize the matching groups (0.077542ms)
✔ negation and conditional language remain fail-closed (0.079584ms)
✔ Chinese discussion of subagents does not authorize delegation (0.061125ms)
✔ gateway intent shares the same negation policy (0.196542ms)
✔ effective child allowlists never advertise parent-only tools (1.86325ms)
✔ explicit child tools are checked against the final bound registry (602.424375ms)
✔ child binding restores only requested child-safe package tools after parent surface gating (384.848584ms)
✔ child denylist keeps extension and workflow structured tools available (730.913167ms)
✔ resource loading gates project resources and keeps only the pi-intercom package parent-only (1135.802667ms)
✔ alternate standalone cwd only uses explicit saved trust (65.768584ms)
✔ shutdown helper balances hooks and disposal despite errors (0.59925ms)
✔ shutdown helper bounds a stuck hook before disposal (10.973042ms)
✔ shutdown helper bounds abort and surfaces cleanup failure (11.418542ms)
✔ every registered package tool is classified child-safe or excluded (fail-closed drift guard) (132.217417ms)
✔ formats current context occupancy against model capacity (1.636ms)
✔ formats the latest post-compaction usage rather than prior cumulative usage (0.079375ms)
✔ clamps over-capacity and nonsensical token values (0.062209ms)
✔ handles missing usage or capacity without NaN or Infinity (0.049792ms)
✔ one session binds all OpenPI editor layers with one UI write (962.569417ms)
✔ keys read brighter than what they do, and a notice takes the line (2.522833ms)
✔ the frame is padded to an exact height and never exceeds its width (0.499625ms)
✔ pre-styled labels and metas are passed through, not repainted (0.262417ms)
✔ a title line stays on one bounded row at any width (0.2555ms)
✔ capability discovery defaults to explicit and accepts adaptive opt-in (0.943875ms)
✔ every one-line footer preset keeps model context left and cwd right (0.638458ms)
✔ the previous canonical footer default migrates without rewriting custom layouts (0.139875ms)
✔ an unreadable config blocks the write and survives untouched (33.69175ms)
✔ a readable or absent config saves normally (63.948459ms)
✔ legacy footerItems migrates and writes back canonical footerLines idempotently (84.4ms)
✔ save and update strip runtime footerItems extras at the writer boundary (59.513375ms)
✔ an update patches the document as it is on disk, not as it was read (96.426791ms)
✔ concurrent Pi processes preserve every read-modify-write update (3164.934667ms)
✔ a killed setup writer releases ownership for the next process (1399.172875ms)
✔ a reused PID does not impersonate the dead lock owner (63.537834ms)
✔ live and uncertain lock owners are never stolen (7689.753292ms)
✔ a stored value that had to be normalized is reported, not hidden (33.438583ms)
✔ legacy footerItems migration remains visible in update reports (26.618208ms)
✔ legacy model-free recaps are explicitly migrated to disabled suggestions (50.979375ms)
✔ the post-edit command is off by default, trimmed, and length-bounded (21.840625ms)
✔ terminal sanitizer removes CSI and terminated or unterminated terminal strings (0.889625ms)
✔ terminal sanitizer removes bidi spoofing controls but preserves shaping joiners (0.225ms)
✔ terminal control detection rejects control-bearing transcript labels (0.604791ms)
✔ zero-byte projections are empty (0.701917ms)
✔ tiny projections never exceed their UTF-8 byte budget (0.370083ms)
✔ the production timeout error names the tool and three-minute limit (0.713792ms)
✔ a hung tool call fails clearly and receives an abort signal (11.662292ms)
✔ parent cancellation still stops the timeout wrapper immediately (0.588834ms)
✔ the guard wraps each definition once and can discover later tools (0.258917ms)
✔ successful and terminating tool results pass through unchanged (0.089209ms)
✔ the timeout is fresh for each tool call, not shared across calls (122.161792ms)
✔ ordinary OpenPI requests match Bare Pi's resident prompt and tools (4144.93325ms)
✔ real Pi adaptive discovery exposes only the gateway (3619.409458ms)
✔ real Pi session starts with the compact OpenPI parent surface (696.631584ms)
✔ real Pi session exposes one stable subagent family when delegate loads (379.876959ms)
✔ real Pi search capability exposes file and git tools as one group (406.299292ms)
✔ separate owner wrappers cannot remove each other's lifecycle tools (270.898375ms)
✔ filesystem ownership does not hide a first-registered foreign namesake (204.903083ms)
✔ owner patch starts from the latest tool list and preserves foreign tools (1.885667ms)
✔ owner patch fails closed on undeclared tools (0.199625ms)
✔ inline ownership fails open without hiding foreign tools (0.18175ms)
✔ an explicitly bound inline owner controls only its declared source (0.08875ms)
✔ owner patch is a no-op when the desired surface is already active (0.090375ms)
✔ catalog defines the compact parent entry surface and every managed name once (0.075ms)
✔ an unloaded capability remembers lifecycle state without exposing its tools (0.111083ms)
✔ visibility catalog and child-session policy classify the same package tools (0.063666ms)
✔ following transcript stays pinned as rows append (0.668125ms)
✔ manual upward scrolling preserves the absolute reading anchor (0.131334ms)
✔ reaching the bottom or explicitly ending restores follow (0.081667ms)
✔ paused viewport clamps safely across shrink and resize without resuming (0.059625ms)
▶ worktreeSlug
  ✔ keeps a readable name and drops everything unsafe (0.597375ms)
  ✔ cannot escape its directory or be read as a git option (0.102542ms)
  ✔ falls back rather than producing an empty segment (0.046791ms)
  ✔ a slug is never unique enough to identify a worktree by itself (0.079875ms)
  ✔ mid-string dots collapse, since git rejects a ref containing .. (0.049667ms)
✔ worktreeSlug (1.511625ms)
▶ worktree lifecycle
  ✔ creates an isolated checkout that the parent status ignores (411.525ms)
  ✔ the child's own status is clean, so node_modules is never committed (593.739625ms)
  ✔ keeps a worktree that still holds uncommitted work (496.30525ms)
  ✔ preserves ignored-only output instead of silently deleting it (401.025208ms)
  ✔ preserves a clean detached checkout instead of guessing it is empty (365.710375ms)
  ✔ preserves a detached HEAD that contains child commits (436.601291ms)
  ✔ preserves a clean branch whose history moved behind its baseline (565.632917ms)
  ✔ a cleanup deadline preserves the checkout before any mutation (235.726292ms)
  ✔ preserves when the immutable base cannot be inspected (393.178959ms)
  ✔ reclaims a committed worktree and keeps the commits on its branch (493.801458ms)
  ✔ the same agent title can be isolated again after one commits (514.07225ms)
  ✔ a productive child is not judged empty because the parent moved (354.160417ms)
  ✔ a child that commits on its own branch keeps it and is reported (334.30025ms)
  ✔ isolates concurrent children from each other (336.84925ms)
  ✔ nests from inside a worktree instead of burying itself (422.348209ms)
✔ worktree lifecycle (6683.9715ms)
▶ worktree outside a repository
  ✔ reports the reason instead of throwing (42.093416ms)
✔ worktree outside a repository (42.237333ms)
✔ built-in roles have exact capability boundaries and no fixed model or effort defaults (1.467167ms)
✔ a valid agent type parses into prompt, tools, model, and effort (3.635583ms)
✔ omitting tools inherits the normal tool set rather than emptying it (0.260708ms)
✔ malformed agent types are reported, never thrown (0.979458ms)
✔ an unrecognized tool name is reported but still applied (0.251208ms)
✔ an over-long body is rejected rather than silently truncated (0.144542ms)
✔ agent type files and model hints are bounded before use (4.676958ms)
✔ an untrusted project contributes no agent types (141.793709ms)
✔ a project agent type overrides the global one of the same name (62.429458ms)
✔ a missing agents directory is normal, and one bad file does not sink the rest (28.569292ms)
✔ an unreadable precedence layer blocks every broader fallback (16.845708ms)
✔ model selection keeps parent inheritance until an override exists (0.198666ms)
✔ agent-type diagnostics strip terminal control sequences (0.057041ms)
✔ childToolPolicy without an allowlist is unchanged (0.064833ms)
✔ an agent-type allowlist narrows and cannot re-enable an excluded tool (0.053666ms)
✔ misspelled frontmatter restriction keys reject the type (0.260084ms)
✔ a malformed higher-precedence role blocks broader fallback (39.156833ms)
✔ a misnamed override blocks its safely declared role (35.648584ms)
✔ every key the parser reads is accepted without a warning (0.6495ms)
✔ naming a parent-only tool says so instead of calling it a typo (0.180541ms)
✔ a symlinked agent type is discovered like a real file (39.019125ms)
✔ a symlink pointing at a directory is still skipped (101.547458ms)
✔ deriveBtwTitle uses the first non-empty line and bounds the title (0.763791ms)
✔ only model-origin snapshots are visible to model-facing tools (0.434083ms)
✔ watermarks survive reload and retain independent model and btw sequences (8.486334ms)
✔ restore takes the high water when concurrent completions arrive out of order (0.11125ms)
✔ legacy tool and result entries migrate without a watermark (0.09025ms)
✔ a fork restores only ids visible on its selected branch (0.080792ms)
✔ malformed and unrelated entries cannot advance either sequence (0.0805ms)
✔ subagent results render before the hidden wake-up message (1.551625ms)
✔ automatic result projection keeps both ends and persists the exact final answer (0.336167ms)
✔ automatic result delivery shrinks a batch against authoritative parent headroom (0.140667ms)
✔ automatic result wrappers and projections stay inside the shared batch cap (3.253ms)
✔ the visible subagent result entry renders the completed report (18.84525ms)
✔ session start preserves the complete registered subagent family (0.884125ms)
✔ the complete subagent family fails closed before the first spawn (4.85825ms)
✔ session_start re-registers agent types for its cwd and live trust decision (34.293084ms)
✔ stub subagent completes and delivers a final result (597.495792ms)
✔ FAIL: prompts settle as errors; unconsumed settles are delivered (312.678458ms)
✔ cancel interrupts a running stub subagent (27.8735ms)
✔ spawn origin propagates to ids, snapshots, and settlement (45.128125ms)
✔ restored session counters continue both id sequences without reuse (4.596542ms)
✔ by-the-way sessions run in their own pool, separate from the model pool (8.447917ms)
✔ the concurrency cap rejects a fifth running subagent (6.332625ms)
✔ a refused spawn releases its concurrency reservation (1.312375ms)
✔ idle restarts respect the concurrency cap (620.997916ms)
✔ restarting a settled subagent settles again so its result re-delivers (1245.125375ms)
✔ send steers an idle subagent into another turn (1193.564041ms)
✔ cancel issued in the restart window stops the new run (581.171167ms)
✔ a wait issued alongside a restart does not return the stale run (1144.22925ms)
✔ a run with no first response is settled by the watchdog and frees its slot (800.188209ms)
✔ a first response clears the watchdog so slower runs are not killed (571.79125ms)
✔ a restart whose run never starts is settled by the watchdog (1326.331334ms)
✔ subagent titles are sanitized and bounded at ingress (0.926667ms)
✔ strip selection prefers newest running, then newest unread settled (0.168834ms)
✔ subagent strip matches Workflow's bounded one-line affordance (13.8625ms)
✔ a lone subagent needs no count: glyph and name carry the state (1.038875ms)
✔ several active subagents aggregate instead of naming just one (0.325208ms)
✔ the pi backend refuses to spawn without the parent's model registry (2.722125ms)
✔ an inherited model that disappeared fails instead of selecting an SDK default (0.775083ms)
✔ an unresolvable model hint fails before a child session exists (0.478083ms)
✔ direct subagents reject a missing declared tool before prompting (2624.970875ms)
✔ a bare id carried by several providers demands qualification (0.354125ms)
✔ the generated agent_type schema exposes a compact, enforced role index (1.543167ms)
✔ reasoning guidance prioritizes the user and task difficulty without fixing a built-in level (0.183542ms)
✔ an explicit user-selected reasoning level remains available (0.435458ms)
✔ the default spawn surface stays within its resident budget (0.202333ms)
✔ the maximum legal roster keeps every enum value while bounding summaries (1.942833ms)
✔ role summaries are deterministic and truncate UTF-8 without splitting it (2.766416ms)
✔ the spawn description derives its concurrency cap from the manager (0.08025ms)
✔ the spawn schema keeps isolation compact while the Skill carries its tradeoffs (15.411709ms)
✔ a planning child reports its effective tools without an agent type (0.213833ms)
✔ a spawned isolated child reports the branch its work will land on (0.121209ms)
✔ interactive spawn guidance releases the turn instead of waiting on dependent work (0.153458ms)
✔ blocking wait is reserved for an explicit synchronous contract (0.101208ms)
✔ short results pass through without creating an artifact (2.445542ms)
✔ byte truncation keeps head and tail and points to the exact artifact (0.435833ms)
✔ the complete projection stays within its byte budget (0.580459ms)
✔ projection budgets remain hard caps across UTF-8 sizes (2.920667ms)
✔ line truncation keeps both ends even below the byte ceiling (0.162583ms)
✔ a long UTF-8 line keeps valid characters at both ends (0.177375ms)
✔ artifact failure is explicit and never advertises a false path (0.11775ms)
✔ content-addressed artifacts are exact, private, and reusable (23.505459ms)
✔ artifact persistence refuses a symlinked cache component (6.064417ms)
✔ artifact persistence refuses an existing file with the wrong content (15.974417ms)
✔ unknown context usage falls back to the deterministic static caps (1.5635ms)
✔ valid parent headroom dynamically shrinks a result batch (0.084458ms)
✔ dynamic headroom includes fixed wrapper metadata (0.077167ms)
✔ short results yield their unused share to longer siblings (0.071125ms)
✔ dynamic batches preserve a readable floor for every result (0.068875ms)
✔ invalid or stale-looking usage never narrows the static fallback (0.072833ms)
✔ empty batches allocate nothing (0.048666ms)
✔ large batches keep their readable floors inside the static batch cap (0.072708ms)
✔ a result consumed by a later wait is not delivered (1.971834ms)
✔ an idle child settlement wakes the parent immediately (0.135ms)
✔ a busy child settlement wakes when the parent settles (0.073792ms)
✔ the parent boundary still delivers if another extension started a run (0.389042ms)
✔ busy results batch in settlement order into one parent wake-up (0.102041ms)
✔ the child-settled and parent-settled edges cannot double deliver (0.075666ms)
✔ a child settling during the wake turn waits for its boundary (0.06725ms)
✔ a synchronous delivery failure restores the batch in order (0.243542ms)
✔ compact automatic results keep the report behind expansion (451.111458ms)
✔ picker and takeover display text cannot inject terminal controls (0.927292ms)
✔ dashboard selection follows its subagent id and falls back by row (1.24275ms)
✔ dashboard box height follows its subagents and retains the old maximum (3.20325ms)
✔ dashboard reserves a more row and shows each visible subagent (0.408208ms)
✔ dashboard uses one status glyph and no flickering live tool (10.439542ms)
✔ chrome rows stay width-bounded and takeover uses three rules (1.364417ms)
✔ takeover is a read-only child page without a message editor (0.763458ms)
✔ takeover scroll indicator lives in its rule without changing overlay height (12.199792ms)
✔ takeover pauses on an absolute reading anchor and resumes at the end (2.211209ms)
✔ transcript sanitization strips terminal control sequences before rendering (0.99075ms)
✔ extension tool calls use their Pi-native renderer instead of raw JSON (2.645833ms)
✔ takeover transcript renders finalized and live assistant Markdown within its width (13.888041ms)
✔ shared transcript renders fenced code and CJK deterministically at narrow widths (17.349125ms)
✔ thinking renders Markdown but preserves redaction (0.95825ms)
✔ tool calls summarize known bounded JSON arguments and safely fall back (0.251417ms)
✔ tool argument summaries relativize paths inside the child cwd (0.157542ms)
✔ tool call and output lines drop the child cwd prefix (0.856292ms)
✔ pending tool calls use the parent activity verbs (0.510833ms)
✔ settled tools use one semantic activity row (0.168416ms)
✔ parallel tool results reuse their earlier command lines (0.137417ms)
✔ tool errors and empty results use status glyphs (7.256041ms)
✔ a running tool becomes settled without reflowing (0.230542ms)
✔ the spinner advances between frames instead of freezing in the cache (0.134833ms)
✔ empty live assistant buffers do not hide live tools or Pi-style queued messages (0.180375ms)
✔ cached items are keyed by width and by tool phase (0.972625ms)
✔ spinnerFrame is deterministic and advances every 120ms (0.051542ms)
✔ tool rendering keeps every line within a narrow width (0.201083ms)
✔ cached finalized transcript output is rebuilt after invalidation (0.2945ms)
✔ wait result preview shows status only and keeps full output behind expand (0.972417ms)
✔ wait result preview bounds status rows across large fan-out (0.090458ms)
✔ setup defaults to disabled next-action suggestions (2.163083ms)
✔ setup status appends optional integration inventory only when supplied (0.1575ms)
✔ setup config accepts suggestion models and migrates the recap key (0.696083ms)
✔ subagent role models parse partially and drop malformed persisted entries (0.1375ms)
✔ workflow limits default safely and accept configured fan-out (0.097792ms)
✔ UI defaults to a compact header and one-line plain footer (0.11325ms)
✔ legacy footerItems migrates onto the default one-line skeleton (0.240416ms)
✔ empty legacy footerItems falls back to the default layout (0.096791ms)
✔ optional-only legacy footerItems remains visible after migration (0.194459ms)
✔ normalizeFooterLines drops unknowns, duplicates, extra flex, and empty rows (0.134291ms)
✔ footerLines is the source of truth when both legacy fields exist (0.105125ms)
✔ presets map to style and lines (0.05475ms)
✔ applyFooterConfig order is current → preset → style/lines; items conflicts with lines (0.238959ms)
✔ registers only bounded suggestion lifecycle hooks and no history renderer (1.363666ms)
✔ parses one bounded one-line suggestion and strips terminal controls (2.91475ms)
✔ rejects invisible output and strips direction controls (0.087291ms)
✔ accepts null and extracts fenced or wrapped JSON (0.106833ms)
✔ rejects malformed or over-specified responses (0.251125ms)
✔ omits reasoning only when configured off (0.649625ms)
✔ prompt predicts user input rather than recapping the run (0.145125ms)
✔ run boundaries replace stale starts and settle exactly once (1.878667ms)
✔ run slicing starts after the before_agent_start leaf (3.005541ms)
✔ transcript omits thinking, images, and recap entries while redacting tool data (1.644459ms)
✔ transcript enforces per-result and total byte caps (1.334042ms)
✔ latest-wins state rejects stale or non-empty-editor offers (0.75325ms)
✔ renders the ghost on the first row with reserved IME preedit cells (2.284583ms)
✔ Right accepts into an empty editor without submitting (2.727084ms)
✔ an invisible suggestion never steals Right from a custom editor (0.251291ms)
✔ an editor too narrow to reserve an IME cell does not steal Right (0.090417ms)
✔ supports a custom editor cursor with a selective reverse reset (0.104042ms)
✔ any other editor input cancels then passes through (0.069459ms)
✔ focused management strips keep Right navigation precedence (0.404ms)
✔ task lifecycle tools appear only after task state exists (3.256209ms)
✔ persistent task widget restores, updates, and expands all tasks (0.883916ms)
✔ mounted task widget renders live task state without another UI write (2.400958ms)
✔ four tracked tasks refresh after every explicit progress transition (2.670167ms)
✔ task tools teach immediate reconciliation without inferring completion signals (0.82275ms)
✔ task panel commands report actual visibility and conflicts block the shortcut (1.099416ms)
✔ detects foreign Todo/plan tools and reports their source (0.093166ms)
✔ injects one transient block into the last user message (0.199208ms)
✔ escapes task-context delimiters inside projected content (0.0685ms)
✔ normalizes string content and skips when no user message exists (0.104292ms)
✔ blocked task tools render their refusal instead of crashing (0.392083ms)
✔ add, update, list, and stable allocation (3.525625ms)
✔ completed batches reset immediately and the next batch restarts at T1 (0.520041ms)
✔ all status transitions are allowed and multiple items may be in progress (0.719459ms)
✔ status entry requires a fresh note and stale notes clear only on leaving (0.441625ms)
✔ strict snapshot validation rejects unknown fields, bad notes, duplicate IDs, limits, and bytes (3.85425ms)
✔ batch and item caps are enforced (0.278584ms)
✔ oversized candidate is rejected before mutable task-state commit (1.472ms)
✔ restore chooses highest revision regardless of position and later entry wins ties (3.453375ms)
✔ foreign session entries are ignored by restore (0.12225ms)
✔ malformed winner and later malformed or unknown entries fail closed (0.327875ms)
✔ restore closes legacy terminal-only batches and restarts their IDs (0.347333ms)
✔ reload after a legacy completed batch keeps the new batch numbering local (0.296375ms)
✔ restore high-water nextId prevents ID rewind inside an active batch (0.247125ms)
✔ concurrent sibling completion order survives reload by monotonic revision (0.137208ms)
✔ branch slices model new, resume, tree, fork, and context-pivot semantics (0.215ms)
✔ projection is empty without actionable work, prioritized, advisory, and bounded (0.928625ms)
✔ render list supports combined id/status filters and character bounds (0.125875ms)
✔ apply functions are synchronous and returned state cannot mutate internal state (0.094583ms)
✔ no-op updates are idempotent without advancing the revision (0.070709ms)
✔ validation errors have a stable public class (0.083041ms)
✔ ordinary newlines and tabs normalize to a single line (0.107958ms)
✔ renders status, detail, and auditable note labels (9.415541ms)
✔ persistent task widget matches a compact Claude-style task panel (1.627708ms)
✔ task widget defaults to four items and Ctrl+Shift+T view shows all (0.431958ms)
✔ batch closure is visible in compact tool results (3.780792ms)
✔ collapsed tool results remain bounded (0.251667ms)
✔ a record without counts shows no census rather than a wrong one (0.107291ms)
✔ a closed batch does not print a census of the batch it just cleared (0.106584ms)
✔ rows are laid out at the width they are rendered at (0.150042ms)
✔ settled subjects are struck through, live ones are not (0.156083ms)
✔ the census line names only the states that are actually occupied (0.144083ms)
✔ the tool result header counts the batch, not the rows it happens to show (0.092667ms)
✔ durations read as seconds, then minutes, then hours (0.632542ms)
✔ default one-line layout leads with model context and ends with cwd (10.139625ms)
✔ legacy footerItems migration preserves the rendered footer (4.449917ms)
✔ powerline emits ANSI256 seams and stays within width (0.516917ms)
✔ powerline-mono uses gray blocks (0.140958ms)
✔ narrow width hides lowest-priority segments before truncating (0.392959ms)
✔ context uses warning and error tones at thresholds (0.097334ms)
✔ uses one Codicon family for model, context, and directory (0.061958ms)
✔ a single status inlines into the first footer line when it fits (0.38075ms)
✔ a single status stays on its own line when it cannot fit (0.134125ms)
✔ operational statuses always append after layout lines (0.132917ms)
✔ statuses remain visible even when metric layout is empty after normalize (0.096166ms)
✔ legacy buildFooterContent still groups selected items (0.127583ms)
✔ fitSegmentsToWidth never empties a previously non-empty side first (0.064625ms)
✔ marks context occupancy unknown instead of guessing a percentage (8.224584ms)
✔ renders only selected footer items (0.787042ms)
✔ always renders operational activity while custom footer is enabled (0.478667ms)
✔ shows the branch but omits changed-file counts (8.267875ms)
✔ config change event reinstalls footer for the active session (0.362417ms)
✔ session_shutdown clears active session so config events do not reinstall (0.192541ms)
✔ model and Git subscriptions recover across session lifecycles without duplicates (3.595041ms)
✔ messages below both thresholds round-trip byte-identical (0.91ms)
✔ a message at exactly the line threshold is unchanged (0.071041ms)
✔ one line over the line threshold folds to a twelve-line preview (0.486833ms)
✔ a message at exactly the character threshold is unchanged (0.07225ms)
✔ one char over the char threshold with one line to hide is left alone (0.061625ms)
✔ long wrapped lines fold by character count, not just line count (0.082833ms)
✔ a char-heavy message with only a few foldable lines stays unchanged (0.083958ms)
✔ folding must hide at least 8 lines to earn its keep (0.069334ms)
✔ a single giant fenced block folds to its first content lines (0.161583ms)
✔ multiple fenced blocks fold per block while prose shares one budget (0.132209ms)
✔ CRLF messages fold without losing their line endings (0.076292ms)
✔ trailing newlines neither fold short messages nor pad the count (0.053875ms)
✔ a fenced message under the thresholds is not folded at all (0.029333ms)
✔ empty and whitespace-only messages are left alone (0.041167ms)
✔ an unterminated fence folds conservatively as plain text (0.0735ms)
✔ many individually short code blocks share one bounded preview budget (0.11925ms)
✔ many truncated code blocks cannot make the folded output longer than the input (0.108375ms)
✔ folding is pure: the input is untouched and repeat calls agree (0.084625ms)
✔ the transformer only folds finalized user messages (0.055458ms)
✔ the marker always states that the model received the full content (0.052583ms)
✔ explicit acceptance accepts only complete evidence (1.178334ms)
✔ rejected, missing, malformed, and missing evidence remain distinct (0.267625ms)
✔ acceptance controls ok without hiding an underlying agent error (0.45475ms)
✔ contract validation rejects duplicate or unsafe identifiers (0.46725ms)
✔ acceptance composes with an existing structured schema (0.7615ms)
✔ artifact transcript keeps the initial prompt, marker, and newest entries (0.929ms)
✔ live artifact persistence includes current agents and transcripts (48.478292ms)
✔ agent result artifacts are complete or fail without leaving a file (33.241583ms)
✔ oversized replay journals fail closed before parsing (14.695125ms)
✔ workflow persistence derives a non-authoritative graph from explicit result refs (6.6365ms)
✔ workflow checkpoints throttle updates and support immediate/final flushes (53.032458ms)
✔ RunController reserves calls synchronously and caps global fanout (19.7975ms)
✔ RunController propagates invocation cancellation without aborting the run (0.484291ms)
✔ RunController enforces call budget and aborts queued tasks (0.37475ms)
✔ RunController double cancel and settle share one terminal result (0.648334ms)
✔ session shutdown aborts active child controllers and bounds completions (11.196208ms)
✔ RunController cleanup timeout is bounded and remains terminal after late completion (12.896083ms)
✔ interactive launch defaults detached while non-delivery hosts wait (1.2105ms)
✔ wait is authoritative and legacy background maps to its inverse (1.748666ms)
✔ conflicting aliases and unsupported detached delivery fail closed (0.251791ms)
✔ wait cancellation does not cancel the underlying completion (0.200959ms)
✔ terminal completion wins when it settles before abort (0.077792ms)
✔ persisted nonterminal invocation facts are projected as uncertain (1.762917ms)
✔ unknown or missing persisted states fail closed without breaking known aliases (0.794ms)
✔ an oversized workflow truncation stub cannot become completed (0.313625ms)
✔ a terminal persisted run cannot retain running agents (0.138041ms)
✔ persisted transcripts retain exact tool call identities (0.135459ms)
✔ stale recovery reconciles the run and every active agent (0.448334ms)
✔ stale pre-V2 runs gain a stable delivery identity while terminal legacy runs do not replay (0.081875ms)
✔ persisted usage is normalized to finite nonnegative numbers (0.09075ms)
✔ saved reports expose the bounded derived graph as observability (0.4065ms)
✔ the dashboard reports the current request, not the session's history (31.515084ms)
✔ restored run directories require a generated safe id (1.302291ms)
✔ a run still executing stays visible even if it began earlier (1.109291ms)
✔ a background run that settles during this request is reported by it (23.875ms)
✔ an ambiguous short suffix cannot open either run in the dashboard (187.798416ms)
✔ direct workflow navigation drills right and returns left through every level (24.940416ms)
✔ Workflow transcript follows, pauses on its top row, and resumes (7.818458ms)
✔ live workflow dashboard repaints on the shared spinner cadence (42.352958ms)
✔ settled child transcript does not repaint for another live run (8.068667ms)
✔ a newly created dashboard report is private (208.654083ms)
✔ overwriting a dashboard report restores private mode atomically (46.576875ms)
✔ agent fields added after a run are not dropped when reading it back (11.885292ms)
✔ narrator lines survive the disk round trip and are re-sanitized (24.583958ms)
✔ foreground run without agents returns the result and persists artifacts (140.54ms)
✔ background runs deliver a follow-up that triggers a turn only when idle (216.079208ms)
✔ a settled launch card does not repaint while its detached run stays active (1038.339333ms)
✔ failed completion delivery remains durable and retries once with the same id (117.041375ms)
✔ a failing script reports the error and records the run as failed (147.497417ms)
✔ agent calls run through the injected session factory and resume replays the journal (1324.651875ms)
✔ oversized authoritative agent results fail without a success record (822.0125ms)
✔ an oversized legacy replay is rejected without a success record (1146.364875ms)
✔ projects explicit result lineage into a stable directed graph (2.338208ms)
✔ diagnoses missing and duplicate references without throwing (0.167667ms)
✔ reports cycles while retaining their persisted edges (0.1145ms)
✔ bounds every collection and reports omitted graph data (0.120792ms)
✔ settled successful output can be referenced within its workflow run (1.579834ms)
✔ resolved entries retain source call identity for graph lineage (0.121208ms)
✔ only settled successful outputs receive references (0.067833ms)
✔ default references are opaque high-entropy tokens (0.240208ms)
✔ token collisions cannot overwrite an earlier conclusion (0.083958ms)
✔ registry rejects invalid configured limits (0.254667ms)
✔ handoff rejects result artifacts that cannot fit its bounded headers (0.543542ms)
✔ resolution rejects requests above the configured reference limit (0.199875ms)
✔ resolution rejects duplicate references (0.195875ms)
✔ resolution rejects unknown and cross-run-like references (0.166209ms)
✔ handoff appends upstream results as untrusted data with stable prompt identity (0.605167ms)
✔ handoffs preserve structured results even when assistant text is present (0.685875ms)
✔ empty text conclusions fall back to safely serialized structured data (0.596125ms)
✔ each text conclusion is capped at 16 KiB without splitting UTF-8 (0.435875ms)
✔ the rendered handoff fairly projects every result inside 48 KiB (2.3515ms)
✔ large fan-out never starves later results by input order (2.202458ms)
✔ the minimum configured handoff budget includes headers and artifacts (1.184917ms)
✔ identity requires a nonblank run id and a positive safe call index (1.482333ms)
✔ a call advances requested -> claimed -> running -> settled success (0.30375ms)
✔ a running call can settle with an error outcome (0.104334ms)
✔ a request may terminate as replayed or rejected instead of running (0.100959ms)
✔ interrupted nonterminal calls settle as uncertain (0.116209ms)
✔ illegal, repeated, and backward-dated transitions fail closed (0.166458ms)
✔ persisted invocation timestamps must remain monotonic (0.096792ms)
✔ record creation and transitions reject malformed durable inputs (0.078208ms)
✔ the call key ignores display options and tracks semantic ones (1.059292ms)
✔ canonical execution ignores equivalent raw model and effort spellings (0.088041ms)
✔ the call key is stable across schema property order (0.061583ms)
✔ repeated identical calls consume their own recorded results in order (0.097416ms)
✔ an empty or absent journal yields a cache that always misses (0.066417ms)
✔ a malformed or old journal degrades to no cache instead of throwing (0.702583ms)
✔ an oversized journal drops the oldest entries and reports how many (84.280292ms)
✔ a bounded journal round-trips back into a working cache (71.681625ms)
✔ a full replay executes no agents and reproduces the original result (172.782083ms)
✔ changing one prompt re-runs only what depends on it (146.035375ms)
✔ replay survives pipeline reordering calls between runs (209.472792ms)
✔ metadata is decoded statically and removed from executable source (3.50375ms)
✔ metadata cannot inject terminal controls into strip or dashboard text (0.459833ms)
✔ export-like text in strings, comments, regexes, and templates is untouched (0.84925ms)
✔ executable and unsupported metadata fail closed (0.576583ms)
✔ active workflow previews and errors cannot inject terminal controls (1.665959ms)
✔ a log line is flattened into one safe terminal row (0.293125ms)
✔ clipping a long line never splits a surrogate pair (0.076333ms)
✔ an empty or whitespace-only log line is dropped, not recorded (0.092792ms)
✔ the log keeps the newest lines and reports how many it dropped (0.149125ms)
✔ a usage reading sums every agent's tokens including cache (0.136333ms)
✔ a usage reading never goes backwards when an agent compacts (0.056792ms)
✔ log() reaches the host in order and phase() stays separate (75.55575ms)
✔ usage() advances as agents settle, so a post-await read sees them (85.127125ms)
✔ usage() exposes the host's resolved call capacity (65.940458ms)
✔ usage() degrades to a zero reading rather than failing the run (74.613792ms)
✔ the usage reading is frozen, so a script cannot fake its own totals (76.659083ms)
✔ an oversized log line is dropped, never fatal to the run (78.249916ms)
✔ an oversized phase title is dropped, never fatal to the run (64.087375ms)
✔ a log flood is capped instead of killing the child with OOM (167.644083ms)
✔ a stage that throws says why instead of leaving a bare null (70.569667ms)
✔ a stage mutating its own input array cannot manufacture nulls (97.217583ms)
✔ Down focuses an available workflow only from an empty editor (1.561166ms)
✔ the focused strip opens right, backs out left, and never traps typing (0.427584ms)
✔ the editor wrapper owns strip keys without changing normal editor input (0.783625ms)
✔ workflow strip sanitizes restored or legacy metadata defensively (2.82675ms)
✔ the workflow strip stays one line, bounded, and exposes its navigation hint (11.498292ms)
✔ workflow strip repaints on the shared spinner cadence (0.897291ms)
✔ reuses one in-memory SessionManager for an operator identity (4.079ms)
✔ serializes concurrent activations of the same key (0.969416ms)
✔ cancels queued work before invoking it (0.479083ms)
✔ continues an operator queue after a rejected activation (0.19675ms)
✔ allows different operator keys to activate concurrently (0.145292ms)
✔ close rejects new work and waits for accepted queues (0.159292ms)
✔ freezes each key to its first execution fingerprint and cwd (0.168208ms)
✔ normalizes valid operator keys without changing their identity (0.070292ms)
✔ rejects unsafe operator keys (0.18225ms)
✔ background follow-up uses a sentence lead-in, not the old bracket form (0.854125ms)
✔ completion batches share one bounded fair projection budget (9.162541ms)
✔ uncertain agents are never described as settled failures (0.17525ms)
✔ launch result advertises the model-facing lifecycle tools (0.070334ms)
✔ lifecycle tool descriptions state their scope and non-blocking nature (0.065875ms)
✔ result message names where isolated work ended up (0.149667ms)
✔ result message surfaces explicit acceptance state (0.069458ms)
✔ result message stays quiet when nothing was isolated (0.050583ms)
✔ workflow status text cannot replay terminal controls from artifacts (0.112125ms)
✔ result message carries the script's narration and what it dropped (0.143416ms)
✔ model projection follows parent headroom and preserves head, tail, and recovery (1.9565ms)
✔ the resident workflow prompt stays compact while the Skill carries the full guide (80.682541ms)
✔ workflow tool errors with malformed details fall back to plain text (1.876333ms)
✔ running workflow cards request and render each shared spinner frame (14.764166ms)
✔ running workflow result messages let the glyph carry the state (0.72275ms)
✔ only provably read-only agent calls are replay-safe (0.767958ms)
✔ replay filesystem boundary permits repo-local observers and rejects escaping paths (225.99275ms)
✔ replay filesystem boundary fails closed for symlinks and uncertain paths (274.636958ms)
✔ replay filesystem boundary rejects path swaps during async execution (454.858292ms)
✔ replay filesystem boundary rejects unknown observer implementations (207.741666ms)
✔ workspace guard rejects replay and journaling across unsafe overlap (0.150791ms)
✔ process-wide guard coordinates overlapping calls from different workflow runs (0.100708ms)
✔ canonical cwd and repository state participate in replay identity (1045.667125ms)
✔ loaded resource content and trust participate in replay identity (410.343625ms)
✔ ignored observable files disable replay rather than returning stale output (214.005458ms)
✔ tracked and untracked symlinks disable replay (407.993166ms)
✔ fingerprint inability fails closed (167.426292ms)
✔ failed delivery stays pending and retries with the same per-run id (1.7725ms)
✔ partial batch receipts retry only unacknowledged runs (0.179667ms)
✔ stale held inline completion restores as pending (0.080875ms)
✔ a receipt persistence failure retains the same delivery for at-least-once recovery (0.14525ms)
✔ a completion queued during a failed flush is drained without another lifecycle event (0.194542ms)
✔ a persistence exception cannot drop later envelopes from a failed batch (0.163375ms)
✔ a persistence exception cannot drop later unacknowledged receipts (0.143959ms)
✔ a held-inline restore persistence failure does not block the final idle flush (0.126083ms)
✔ a run id is resolved by exact match or by hex suffix (1.837ms)
✔ an ambiguous short suffix cannot resume either matching run (0.318375ms)
✔ a traversing run id cannot reach a journal outside the runs directory (0.335625ms)
✔ runAgent uses a caller-supplied in-memory session manager (5.26325ms)
✔ completed parallel tool calls pair lifecycle timings with calls and results (1.516208ms)
✔ in-flight aborted tool calls retain start timing without completion (0.306542ms)
✔ assistant settlement survives failed compaction and clears after recovery (0.267292ms)
✔ schema-less agents fail when prompt resolves without an assistant response (0.612584ms)
✔ schema-less agents accept an empty assistant message_end (0.6175ms)
✔ cancel during prompt owns the session and returns after bounded disposal (13.103875ms)
✔ cancel during a hanging tool ignores late events and progress writers (2.334375ms)
✔ late prompt completion after first-response timeout cannot become success (5.107041ms)
✔ cleanup timeout is surfaced instead of reporting agent success (6.159417ms)
✔ first-response watchdog aborts a silent provider request (11.218ms)
✔ first assistant response disarms the watchdog without limiting the run (21.072791ms)
✔ explicit watchdog cancellation disarms a pending operation (16.000333ms)
✔ structured role children keep their terminating tool without widening capabilities (0.152ms)
✔ workflow agents reject a missing declared tool before prompting (249.546875ms)
✔ workflow child guards enforce the replay filesystem boundary (3.398792ms)
✔ workflow children guard structured, normal, and dynamically registered tools (10.600084ms)
✔ sandbox exposes only workflow capabilities and validates results (95.938958ms)
✔ sandbox result serialization handles cycles and bigint (91.4515ms)
✔ sandbox rejects unawaited agent calls (72.724625ms)
✔ sandbox source cannot escape the host accounting wrapper (66.224917ms)
✔ sandbox VM still rejects non-yielding synchronous code (1067.817375ms)
✔ workflow agent invocations have no per-request wall timer (108.760708ms)
✔ workflow cancellation aborts a pending agent request (81.46825ms)
✔ late agent completion after cancellation is ignored (68.764833ms)
✔ the sandbox budget is the configured budget, not a lower hidden one (148.126125ms)
✔ the sandbox hard cap sits a fixed margin above the configured budget (130.816833ms)
✔ parallel() settles a throwing thunk to null without failing the batch (57.283208ms)
✔ pipeline() advances each item independently, with no barrier between stages (124.621709ms)
✔ pipeline() drops a throwing item to null and skips its remaining stages (56.751666ms)
✔ pipeline() honors the concurrency cap and an empty input (90.450917ms)
✔ pipeline() rejects a non-array input and non-function stages (111.085208ms)
✔ pipeline() with no stages returns the items unchanged (55.740791ms)
✔ a promise handed to workflow code cannot reach the host realm (57.652458ms)
✔ agent_type reaches the host verbatim, per agent call (55.314208ms)
✔ acceptance reaches the host verbatim, per agent call (56.540125ms)
✔ operator and result-reference inputs reach the host verbatim (60.097792ms)
✔ isolation reaches the host verbatim, per agent call (55.601166ms)
✔ replayed calls raise the backstop instead of killing a resumed run (120.1735ms)
✔ complete JSON encoding stops reading arrays at the first hard limit (2.957708ms)
✔ complete JSON encoding charges escaped UTF-8 bytes before reading values (0.112042ms)
✔ complete JSON encoding preserves supported values within its exact cap (0.10275ms)
✔ lossy serialization also stops collection reads at the node limit (1.863042ms)
✔ complete JSON encoding contains object enumeration failures (0.151167ms)
✔ complete JSON encoding charges sparse array slots to the node budget (0.0945ms)
✔ complete JSON encoding rejects non-finite budget configuration (0.217542ms)
✔ oversized object keys fail before their values are read (0.344125ms)
✔ safeStringify handles cycles, bigint, depth, and size (0.165416ms)
✔ atomic writes leave complete readable content (7.878917ms)
✔ run target resolution prefers exact ids and bounds ambiguous or missing errors (1.672375ms)
✔ an ambiguous short suffix cannot stop or inspect either active run (480.582166ms)
✔ session start keeps the complete workflow capability group stable (145.050125ms)
✔ Direct and Workflow adapters render one equivalent conversation body (27.647584ms)
✔ Workflow children use the same native renderer as Direct children (1.206708ms)
✔ old Workflow transcript entries without call ids remain renderable (0.251292ms)
✔ explicit results consume pending calls before legacy id fallback (0.246417ms)
✔ handoff captures tracked binary patch plus untracked and ignored inventories (755.239208ms)
✔ handoff capture failure is explicit and leaves cleanup to the caller (593.002291ms)
✔ confirmed deletion of a pre-existing file proceeds (10.658542ms)
✔ refused deletion of a pre-existing file is blocked (2.181167ms)
✔ cancelling the active turn dismisses confirmation and blocks deletion (2.416ms)
✔ confirmation errors fail closed and do not corrupt a retry (3.212542ms)
✔ a file created through Write can be cleaned up without confirmation (2.325459ms)
✔ settlement clears provenance from the completed agent run (2.839417ms)
✔ r11-style mixed cleanup cannot delete a baseline file (22.943375ms)
✔ session-created scratch can be deleted without confirmation (4.010833ms)
✔ scratch created by the native write tool can be deleted without confirmation (6.068042ms)
✔ native write keeps an overwritten baseline file protected (6.480708ms)
✔ a failed native write never grants scratch deletion authority (5.960875ms)
✔ confirmed deletion of a pre-existing file is allowed (4.608875ms)
✔ ambiguous shell deletion is not misrepresented as protected (3.617625ms)
✔ a later command failure does not erase proven scratch creation (5.507125ms)
✔ a literal mkdir target remains session-owned after a later failure (4.153541ms)
✔ a sandbox wrapper cannot hide a protected cleanup from the guard (4.074417ms)
ℹ tests 880
ℹ suites 3
ℹ pass 880
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 25200.013167

 RUN  v4.1.10 /Users/tushaokun/work/my-pi-setup


 Test Files  1 passed (1)
      Tests  30 passed (30)
   Start at  13:27:51
   Duration  2.93s (transform 70ms, setup 0ms, import 1.08s, tests 1.71s, environment 0ms) on head 035c66d runs 871/871 (plus Vitest 30/30). A clean local integration with current main, including #147's four startup-cancellation regressions, runs 875/875 (plus Vitest 30/30), and Checked 295 files in 163ms. No fixes applied.
Found 2 errors. passes. GitHub also currently marks the branch behind/conflicting. Please update the branch and replace the validation receipt with the counts from the updated head so the migration's test-count evidence is exact. I also verified the updated tree's npm pack dry-run contains neither  nor the test-only subagent stub.

tt-a1i (dismissed):
Thanks for the thorough test-layout cleanup. The implementation itself looks sound, and I found no code-level blocker. One evidence issue needs correction before approval: the PR reports the Node suite as 881/881, but the current head 035c66d runs 871/871, plus Vitest 30/30. A clean local integration with current main, including the four startup-cancellation regressions from PR 147, runs 875/875, plus Vitest 30/30, and bun run check passes. GitHub currently marks the branch behind/conflicting. Please update the branch and replace the validation receipt with counts from the updated head so the migration evidence is exact. I also verified that the integrated tree npm pack dry-run contains neither the tests directory nor the test-only subagent stub.

tt-a1i (approved):
Reviewed the updated head. The test-only Subagent stub now lives under tests/support, its consumer import and ownership guard are correct, and the affected documentation matches the top-level test layout. Local validation passed: bun run check, Node 875/875, Vitest 30/30, package dry-run with no test-only assets, and git diff --check.

PR #201, opened 2026-08-26
fix(workspace-cleanup-guard): block opaque rm targets

### LINKED ISSUE #172: workspace-cleanup-guard: 动态 rm target 可绕过已有文件删除确认
## 问题

Workspace cleanup guard 能保护字面量 `rm` 目标，但对动态/不透明目标仅设置 `opaqueDestructiveCommand: true`，仍然返回 `kind: "allow"`。因此变量、glob、command substitution 等常见 shell 写法可以绕过已有文件删除确认。

## 最小复现

工作区预先存在 `keep.txt`：

```sh
target="keep.txt"; rm "$target"
```

`inspectShell()` 无法把 `$target` 解析为字面路径，于是标记 opaque；但 `before()` 仍允许命令执行。现有测试甚至锁定了该行为：

```ts
assert.equal(decision.kind, "allow")
assert.equal(decision.opaqueDestructiveCommand, true)
```

## 根因

- `extensions/workspace-cleanup-guard/workspace-provenance.ts:193-201`：动态 `rm` target 被标记 opaque；
- `extensions/workspace-cleanup-guard/workspace-provenance.ts:299-302`：opaque 状态没有参与 allow/block 决策；
- `extensions/workspace-cleanup-guard/index.ts:38-40`：调用方只阻止 `kind === "block"`；
- `extensions/workspace-cleanup-guard/workspace-provenance.test.ts:195-207`：回归测试要求动态删除被放行。

这违反 README 对 pre-existing workspace paths 删除保护的承诺。攻击不需要复杂 shell：单个变量即可绕过。

## 预期行为

无法证明删除目标时 fail closed：

- 阻止命令执行；
- 不调用针对“已知路径”的确认 UI，因为此时无法准确列出目标；
- 给出可操作说明：改用字面量 workspace-relative path，让 guard 能验证其是否为 session-created scratch 或已知 pre-existing path；
- 不向 pending effects 写入未验证删除状态。

未来若要支持动态 shell，可另行设计真实 shell AST/受控展开和 reviewed confirmation；当前不能把 opaque 当作 informational flag 后继续执行。

## 验收标准

1. `target="keep.txt"; rm "$target"` 返回 `kind: "block"`；
2. 预先存在的 `keep.txt` 保持不变；
3. unknown target 不触发带错误路径列表的 confirmation；
4. 字面量 baseline 删除仍走原生确认；
5. session-created scratch 的字面量删除仍直接允许；
6. nested shell 中的 opaque `rm` 同样 fail closed；
7. `bun run check && bun run test` 全部通过。

## 严重程度

P0 data-loss boundary：可直接绕过项目承诺的已有文件删除保护。

sowen1023:
## Summary

- keep direct literal `rm` on the existing provenance and native confirmation path
- fail closed when source-visible `rm` appears in compound, nested, forwarded, or dynamically targeted commands that cannot be verified safely
- preserve recognizable non-executable `rm` text in ordinary arguments, Bash comments, and a standalone non-expanding heredoc
- document the guard as a narrow UX boundary rather than a filesystem sandbox or a guarantee for runtime-generated command names

## Tests

- `bun run check`
- `node --test --experimental-strip-types tests/extensions/workspace-cleanup-guard/workspace-provenance.test.ts tests/extensions/workspace-cleanup-guard/index.test.ts` (22/22)
- `bun run test` (881/881 Node tests; 30/30 Vitest tests)
- `npm pack --dry-run --json --ignore-scripts`

Closes #172


extensions/workspace-cleanup-guard/workspace-provenance.ts
```diff
@@ -79,66 +183,104 @@ function splitShellSegments(command: string) {
...
+  if (!token || token.value === "--" || token.dynamicPath) return undefined;
+  return token.value;
 }
 
 function inspectShell(command: string): ShellInspection {
   const creations = new Set<string>();
   const removals = new Set<string>();
   let opaqueDestructiveCommand = false;
-  for (const segment of splitShellSegments(command)) {
+  for (const segment of splitShellSegments(withoutHeredocBodies(command))) {
```
tt-a1i: This still assumes that each segment has one executable at tokens[0], but splitShellSegments does not split a single pipe or model compound shell command positions, and this path does not unwrap the command builtin. I reproduced all of the following as decision.kind === "allow":

- target="keep.txt"; printf x | rm "$target"
- target="keep.txt"; command rm "$target"
- target="keep.txt"; (rm "$target")
- target="keep.txt"; if true; then rm "$target"; fi
- target="keep.txt"; { rm "$target"; }

Each form executes rm in the shell, so the protected target can still be deleted. Please recognize these executable command positions or fail closed when the parser cannot prove that a destructive form is safe, and retain them as regression tests.

tt-a1i (changes requested):
The direct dynamic-target cases added here behave as intended, and the existing checks pass. However, the fail-closed boundary is still bypassable when rm is not the first executable token in a parsed segment. See the inline finding for reproduced cases.

Local verification on this head:
- bun run check: passed
- workspace-cleanup-guard tests: 20/20 passed
- full Node suite: 877/877 passed
- Vitest: 30/30 passed

Please add regression coverage for the reproduced command positions and make unsupported destructive forms fail closed before execution.

tt-a1i:
Thanks for tightening the direct dynamic-target cases. I found one remaining blocking gap in command-position handling: pipeline, command-wrapper, subshell, if-body, and brace-group forms can still execute rm while the guard returns allow. I included the exact reproduced commands in the formal review. The existing local checks all pass, so the requested change is specifically to close these uncovered paths and add regressions.

extensions/workspace-cleanup-guard/workspace-provenance.ts
sowen1023: resolved

extensions/workspace-cleanup-guard/workspace-provenance.ts
```diff
@@ -148,25 +358,29 @@ function inspectShell(command: string): ShellInspection {
...
     }
     for (let index = 0; index < tokens.length; index += 1) {
-      if (tokens[index] !== ">") continue;
+      if (tokens[index]?.value !== ">") continue;
       const target = literalPath(tokens[index + 1]);
       if (target) creations.add(target);
     }
 
-    const executable = tokens[0]?.split("/").at(-1);
+    const commandTokens = executableTokens(tokens);
```
tt-a1i: `executableTokens()` only unwraps a small set of top-level prefixes, so nested executable contexts are still treated as non-destructive segments. For example, command substitution, `sh -c`, `eval`, `case`, function bodies, `time`, and an expanding heredoc all returned `allow` in isolated probes and then deleted the pre-existing target when executed with Bash. This needs a fail-closed boundary for unsupported shell execution, with regressions for the reproduced forms, rather than another finite prefix list.

tt-a1i (changes requested):
The five command positions from the previous review are now covered, but the fail-closed boundary remains bypassable through other executable shell contexts.

I ran isolated probes against `d410559`. Each command below returned `decision.kind === "allow"`; executing it with Bash then deleted a pre-existing `keep.txt`:

```sh
target="keep.txt"; echo "$(rm "$target")"
target="keep.txt"; output=$(rm "$target")
target="keep.txt"; echo `rm "$target"`
target="keep.txt"; cat <(rm "$target")
sh -c 'rm "$1"' sh keep.txt
target="keep.txt"; eval 'rm "$target"'
target="keep.txt"; case x in x) rm "$target";; esac
target="keep.txt"; f(){ rm "$target"; }; f
target="keep.txt"; time rm "$target"
target="keep.txt"
cat <<EOF
$(rm "$target")
EOF
```

This still violates Issue #172's requirements that a pre-existing target remain unchanged and that nested opaque `rm` forms fail closed. Please address the unsupported shell-execution class rather than only the currently enumerated prefixes, and retain the reproduced forms as regressions.

Local verification on this head:

- `bun run check`: passed
- workspace-cleanup-guard tests: 22/22 passed
- full Node suite: 879/879 passed
- Vitest: 30/30 passed
- isolated bypass probes: 0/10 blocked; all 10 reached deletion

tt-a1i:
Thanks for addressing the five command positions from the previous review. I re-tested `d410559`; the formal review documents ten remaining shell forms that still return `allow` and delete a pre-existing target. The existing checks are green, so the remaining blocker is specifically the uncovered fail-closed execution boundary.

extensions/workspace-cleanup-guard/workspace-provenance.ts
```diff
@@ -31,24 +31,148 @@ interface ShellInspection {
...
+}
+
+interface Heredoc {
+  delimiter: string;
+  stripTabs: boolean;
+}
+
+const DYNAMIC_UNQUOTED_PATH_CHARACTER = /[*?[$`{}]/u;
+const SHELL_ASSIGNMENT = /^[A-Za-z_][A-Za-z0-9_]*=/u;
+const COMMAND_POSITION_PREFIXES = new Set([
```
tt-a1i: This finite prefix registry is becoming an implicit shell grammar: every missing executable position is another deletion bypass. The smaller provable contract is to keep the conservative opaque-`rm` fallback from `main`, block opaque input, and allow only directly verified literal cleanup; callers can retry complex cleanup as a standalone command. If arbitrary Bash composition must remain accepted, that requires a reviewed AST owner rather than extending this list.

tt-a1i (changes requested):
Looking at this specifically as an ownership and simplification problem, I think the current direction is carrying the wrong contract.

### Ranked simplification candidate: narrow the accepted cleanup grammar

- **Candidate:** retain the provenance/confirmation state machine, but remove the attempt to recognize an expanding set of Bash command positions. Preserve the conservative `main` fallback that classifies an unproven `rm` occurrence as opaque, make opaque destructive input block, and only allow deletion when a direct literal workspace-relative target is proved.
- **Burden:** this PR adds 221 net production lines and 204 net test lines, including heredoc parsing, quote/escape provenance, subshell splitting, command-prefix recognition, and wrapper unwrapping. Those pieces collectively become a partial second shell grammar.
- **Reachability:** the parser has one production owner: the Bash `tool_call` guard. The `origins` and `pending` maps, filesystem observations, native confirmation, result settlement, and `agent_settled` reset are separate real contracts and should stay.
- **Evidence:** despite the added grammar, ten current-head probes still return `allow` and delete a pre-existing target through command/process substitution, backticks, `sh -c`, `eval`, `case`, a function body, `time`, and an expanding heredoc. Adding more prefixes does not close the class.
- **Consequence:** a complex or harmless command containing an unproven `rm` may be blocked and must be retried as a standalone literal cleanup. That is a visible, recoverable restriction. If preserving arbitrary Bash composition is a product requirement, it needs a reviewed shell AST boundary; there is no existing parser dependency in this package that currently owns those semantics.
- **Proof:** direct literal baseline deletion still opens native confirmation; direct literal session-created scratch cleanup still proceeds; every dynamic or nested deletion probe blocks before execution and opens no false-path confirmation; blocked attempts create no pending effects.

I would not simplify away the guard, provenance map, pending-effects settlement, confirmation, or reset lifecycle: each protects the documented pre-existing-file boundary. I also would not add another set of command-position cases, because that retains the same ownerless grammar obligation without making the safety claim provable.

Please reduce this PR to the narrow fail-closed contract required by Issue #172, or explicitly choose and test a full shell-parser dependency if arbitrary shell composition must remain supported.

tt-a1i:
I added a focused simplification review. The data-loss guard and provenance lifecycle should remain; the removable burden is the growing partial Bash grammar. The review proposes a narrower fail-closed contract, states its user-visible tradeoff, and gives the checks that should prove the cut.

tt-a1i:
Thanks for the update — this narrower direction is much easier to reason about.

I pushed a small follow-up commit (`9284e48`) to the PR branch after reviewing the updated implementation. It keeps ordinary Bash Pi-native, while tightening the standalone `rm` boundary:

- reconstructs literal `rm` references across quote, escape, and line-continuation forms so wrapped cleanup still blocks before confirmation;
- avoids treating ordinary identifiers such as `rm-notes.txt` or `"r\\m"` as cleanup;
- matches Bash semantics for quoted backslashes, line continuations, and non-ASCII filename characters;
- distinguishes real `../` traversal from valid in-workspace names such as `..foo`.

Validation is green:

- `bun run check`
- `bun run test` — 877 Node tests and 30 Vitest tests passed

The AST dependency remains removed, and the guard stays scoped to the standalone direct-`rm` contract. Please take a look when you have a chance.

tt-a1i (changes requested):
The narrower direct-`rm` grammar is a meaningful improvement, and the previously reported pipeline/subshell/heredoc positions are now covered. However, the new negative text check still leaves the P0 boundary bypassable.

At `workspace-provenance.ts:297`, any command for which `containsRmReference()` cannot reconstruct the literal token `rm` is treated as ordinary Bash and allowed. On current head `422135a`, I reproduced both of these commands:

```sh
$'\x72\x6d' keep.txt
$(printf '\162\155') keep.txt
```

For each probe, `keep.txt` existed before the guard was created, `guard.before()` returned `kind === "allow"`, no confirmation was opened, `/bin/bash -c` exited successfully, and `keep.txt` was deleted.

This is the same unsupported dynamic-execution class that the PR's README contract says should fail closed. Adding more spellings to `containsRmReference()` will not make a negative textual heuristic prove that arbitrary Bash has no deletion effect. Please move the allow/block boundary to something that positively proves the supported grammar, or narrow the product contract and Issue #172 acceptance so it no longer claims this protection.

Local verification on `422135a`:

- `bun run check`: passed
- workspace-cleanup-guard tests: 20/20 passed
- combined Node run with the isolated probe: 877 passed, 1 failed (the reproduced bypass above)
- Vitest: 30/30 passed
- GitHub currently reports the PR as conflicting with latest `main`, with no CI checks attached to this head

This remains blocking because a pre-existing workspace file is actually deleted before any confirmation.

tt-a1i (approved):
Re-reviewed after the source-visible rm boundary was narrowed and the bypass regressions were fixed. Focused tests, full check, and the full suite pass.

PR #220, opened 2026-08-28
fix(tui): stop settled strip redraw timers

### LINKED ISSUE #189: perf(tui): settled Workflow/Subagent strip 仍持续定时请求全局重绘
## 问题
Workflow 与 Direct Subagent 的 below-editor strip 只要可见就启动固定 timer，并无条件 `tui.requestRender()`。即使最新条目已经 settled、glyph 和文本完全静态，Workflow 仍约每 120ms、Subagent 每 500ms 请求全局 render；settled unread 状态会一直保留到用户下一次明确请求，因此空闲界面也可长期重绘。

## 证据
- `extensions/workflows/navigation.ts:41-67`：widget 构造后以 `SPINNER_INTERVAL_MS` 无条件 requestRender；
- `extensions/workflows/index.ts:854-886`：settled run 仍保持 widget 可见；
- `extensions/subagents/navigation.ts:72-96`：每 500ms 无条件 requestRender；
- `extensions/subagents/index.ts:421-443`：settled unread snapshot 仍保持 widget 可见。

与 #96 不同：#96 是 detached Workflow 历史 tool card 动画导致 regular TUI scrollback 重绘；本问题是 below-editor settled strip 的静态 idle polling。

## 期望
- timer 仅在 glyph 真正动态（running）时存在；
- settled/failed/uncertain 静态状态只在状态、focus、width/theme 变化时 render；
- running → settled 必须停止 timer；重新运行再启动；
- fake timer 测试证明 settled widget 多周期 invalidate/requestRender 为 0；
- idle benchmark 记录 60 秒无活动下 requestRender 次数和 CPU，应接近 0。

6iKUN6:
## Problem

Fixes #189. Workflow and Subagent below-editor strips continued calling `requestRender()` while a settled unread result remained visible, so a finished run kept the terminal repainting a line that never changed.

Reviewing the first pass surfaced a second, larger source of the same idle cost: Pi's `setStatus()` requests a global render unconditionally, without diffing the text (`interactive-mode.js` `setExtensionStatus`). The subagent manager notifies on *every* child event (`foldEvent` -> `notify`), and each notify reached `ui.setStatus`. In the TUI that value is always `undefined` — the strip carries the activity instead — so streaming deltas, tool cycles, and usage ramps each repainted the whole terminal to rewrite nothing. The widget-level repaint key could not prevent this: `setStatus` had already requested the render before the widget gate ran.

## Value

Preserves the final result notice and all live feedback without keeping the terminal UI active. Idle `requestRender` calls from a settled strip drop to zero, and a streaming subagent now writes the footer once instead of once per event.

## Approach

- Start a strip refresh timer only while its selected entry is `running`; retain the existing 120 ms Workflow and 500 ms Subagent cadences.
- On settlement, clear the timer and request one final render for the terminal state; restart it if the entry runs again.
- Gate immediate strip refreshes on the selected entry identity and lifecycle status.
- Add `createStatusWriter` in `extensions/shared/activity-status.ts`: it remembers the last value written per UI object and forwards only real changes. Both Subagent and Workflow footer writes route through it, and both reset it on `session_shutdown` (a new session's footer starts empty, so the UI identity is part of the comparison).
- Add `__setSubagentTestBackends`, a test-only backend seam matching the existing `__setWorkflowTestAgentSessionFactory` convention, so extension-level tests can drive a real spawn without a child pi session.

## Validation

- `node --test --experimental-strip-types tests/extensions/workflows/navigation.test.ts tests/extensions/subagents/navigation.test.ts tests/extensions/subagents/status-repaint.test.ts tests/extensions/shared/activity-status.test.ts` — 24 passed, 0 failed.
- `bun run check` — passed; reports the 12 pre-existing Effect warnings in `extensions/file-search/src/binaries.ts`.
- `bun run test` — 937 passed, 0 failed.
- New `tests/extensions/subagents/status-repaint.test.ts` boots the extension with a UI spy and drives a real spawn through the stub backend, covering `manager.notify -> updateStatus -> setStatus -> requestRender`. Verified as a regression guard: reverting the writer makes it fail with 25 status writes instead of 3.
  - TUI: exactly one footer write (`undefined`) across the whole streaming run, and no further widget renders for 900 ms after settlement.
  - Non-TUI: still exactly the three real transitions — empty, `1 running`, `1 done` — so the gate does not swallow genuine changes.
- New `benchmarks/strip-idle-render.ts`, the 60-second idle measurement #189 asked for (`node --experimental-strip-types benchmarks/strip-idle-render.ts --seconds 60`, Darwin arm64, Node 24):

  | phase | renderRequests | cpuMs | cpuPercent |
  | --- | --- | --- | --- |
  | `process-floor` | 0 | 2.5 | 0.004 |
  | `settled-idle-before` | 610 | 355.3 | 0.592 |
  | `settled-idle-after` | 0 | 4.1 | 0.007 |
  | `running-after` | 611 | 265.2 | 0.442 |

  Idle repaints go 610 -> 0 over 60 s, with CPU landing on the bare-process floor; the `running-after` control confirms the spinner still animates at its old ~10.2/s combined cadence. Scope note: the benchmark measures the strips' own timers and render work, not the rest of the frame a real `requestRender()` repaints, so the true idle saving is larger than the delta shown, never smaller.
- Manual Pi TUI `/reload` smoke — run against a locally linked checkout of this branch, covering both strips:
  - Subagent strip (500 ms): spinner animates while running, the elapsed and `% ctx` fields advance, and refreshing them on the spinner tick instead of per event is not perceptible.
  - On settlement the line freezes as `✓ <title>` and **stays on screen** — the unread result notice is preserved.
  - No visible repaint during the idle window that follows.
  - The next user input clears the strip, so the unread/acknowledge semantics are unchanged.
  - Starting another subagent from that settled state makes the spinner animate again, confirming the timer restarts.
  - Workflow strip (120 ms): same lifecycle, same result.
- Manual CPU sampling (`ps -o time=` deltas over 60 s idle, Darwin arm64): with this branch loaded, the process burned 0.28 s, 0.28 s, and 0.12 s of CPU across three idle windows (settled subagent strip visible, settled workflow strip visible, and after a restart cycle). **No real-hardware pre-fix baseline or strip-absent floor was collected**, so these are absolute observations only — they do not by themselves quantify the saving. The CPU-delta claim rests on `benchmarks/strip-idle-render.ts` above, which is an isolated measurement with both a process floor and a pre-fix arm.

## Impact

- User-visible behavior: running strips still animate at the same cadence; settled unread strips remain visible but stop polling for renders. While a subagent runs, the strip's `% ctx` and counts now refresh on the 500 ms spinner tick rather than on every event — the same cadence the glyph already used.
- Runtime/lifecycle: timers are cleared on settlement and disposal; restarting work recreates the timer. The status writer's cache is reset on `session_shutdown`.
- Compatibility/risk: limited to TUI repaint scheduling. Footer text, status rendering, navigation, and result persistence are unchanged; the non-TUI footer assertion above pins that.


extensions/subagents/index.ts
```diff
@@ -422,10 +424,19 @@ export default function (pi: ExtensionAPI) {
...
     if (!ctx || ctx.mode !== "tui") return;
-    const visible = Boolean(stripEntry());
-    if (visible === widgetVisible) return;
+    const entry = stripEntry();
+    const visible = Boolean(entry);
+    const entryKey = subagentStripEntryKey(entry);
+    if (visible === widgetVisible) {
+      if (visible && entryKey !== widgetEntryKey) {
+        widgetEntryKey = entryKey;
+        requestWidgetRender?.();
```
tt-a1i, line 436: Blocking: this repaint gate runs after `updateStatus()` has already called `ui.setStatus("subagents", undefined)`. Pi implements every `setStatus()` call with `requestRender()`, while AssistantDelta, ToolUpdate, UsageChanged, and the other manager events all notify the subscriber. Those events therefore still trigger global renders regardless of this key. Please suppress unchanged TUI status writes (or gate them before calling Pi) and verify the full manager-notify → extension UI path.

tests/extensions/subagents/navigation.test.ts
```diff
@@ -175,3 +207,67 @@ test("several active subagents aggregate instead of naming just one", () => {
...
+      requestRender() {
+        renders += 1;
+      },
+    } as unknown as TUI,
+    theme,
+    new BelowEditorStripState(),
+    () => entry,
+  );
+  try {
+    t.mock.timers.tick(60_000);
```
tt-a1i, line 230: This proves only that the isolated widget timer is idle. It does not exercise `manager.notify → updateStatus → ui.setStatus → requestRender`, so it cannot catch the bypass above. Please add an extension-level UI spy test; separately, Issue #189 asks for a real 60-second requestRender/CPU idle benchmark, which fake timer advancement cannot provide.

tt-a1i (changes requested):
Changes requested. The status-dependent timer lifecycle itself is sound and the targeted tests pass, but the real extension event path still bypasses the new repaint key: every Subagent event reaches `updateStatus()`, which calls `ui.setStatus()` even in TUI mode, and Pi `setStatus()` unconditionally requests a global render. Therefore the PR body claim that streaming, tool, and usage updates no longer cause global strip redraws is not yet true. Please gate or avoid unchanged TUI status writes and cover the complete event path. Issue #189 also explicitly requests a 60-second idle requestRender/CPU benchmark; advancing a fake clock proves timer behavior but supplies no CPU evidence.

agnitum2009 (commented):
Independent second review — local verification plus source-level corroboration of @tt-a1i's blocking points.

**Local verification against c88d8be** (fetched the branch and ran the stated test plan):
- `node --test --experimental-strip-types tests/extensions/workflows/navigation.test.ts tests/extensions/subagents/navigation.test.ts` -> 17/17 pass
- `bun run check` -> clean (only pre-existing effect lint warnings in `extensions/file-search/src/binaries.ts`, unrelated to this PR)
- `bun run test` -> 934/934 pass

So the widget-level timer lifecycle and its fake-timer tests are sound, matching the owner's read.

**Bypass independently confirmed in source**: both extensions subscribe `updateStatus`/`updateIndicator` to every manager event (`extensions/subagents/index.ts:409`), and each calls `ui.setStatus(...)` unconditionally; on the Pi side, `setExtensionStatus()` in `pi-coding-agent/dist/modes/interactive/interactive-mode.js` always calls `this.ui.requestRender()` without diffing the text. So streaming/tool/usage events still force a global render even when the strip is settled — the new entryKey gate stops the widget's own interval but not the event path.

+1 to the outstanding requests: gate unchanged status writes before calling Pi, add an extension-level UI spy test covering manager.notify -> setStatus -> requestRender, and the real 60-second idle requestRender/CPU benchmark from #189.

tt-a1i (dismissed):
Re-reviewed exact head `91ebe4e1bdd2dbb83dc823a4b5bea862a471a3b7`.

The follow-up commit resolves both blockers from the earlier review:

- unchanged Subagent/Workflow status writes are collapsed before Pi's unconditional global render path, with extension-level coverage for the full manager -> setStatus flow;
- the requested real-time idle benchmark is present, while focused tests cover settlement, idle zero-render behavior, and timer restart.

Independent verification:
- focused repaint/navigation tests: 24 passed
- `bun run check`: passed
- `bun run test`: 937 passed
- benchmark smoke: settled idle render requests 9 -> 0 over 1 second; running control remained active
- local merge with current `main@bf6b439`: clean; focused tests and `bun run check` passed on the merge result

No blocking code findings remain on this head. Approval is code-review evidence only: the branch is currently behind `main` and GitHub reports no checks for this head, so required checks still need to run before merge.

6iKUN6:
Rebased onto the latest main; all required checks are green, and the previously requested review threads are resolved. This PR should now be ready to merge. Thank you for the review.

tt-a1i (approved):
Re-reviewed exact head 59d9a7e459b5f7adf8260c25eb6b09f1dbce94d0 against current main efd804dce03687b3082792fb9450e4c07e169ad7. The prior blockers remain resolved: settled strips stop their timers, lifecycle changes receive one final repaint and restart correctly, and unchanged status writes are collapsed before Pi setStatus triggers a global render. The post-approval changes only update the benchmark SubagentSnapshot fixture contract and merge current main. Independent verification: 24 focused repaint/navigation tests passed; bun run check passed; bun run test passed (1009 passed, 1 skipped); the 1-second benchmark smoke measured settled idle render requests 9 -> 0 while the running control remained active. No blocking findings.

PR #224, opened 2026-08-28
perf: skip parent-only workflow extensions in child sessions

### LINKED ISSUE #183: perf(workflows): child 启动会同步 hydrate 全部 foreign workflow artifacts
## 问题
每个 child Session 启动时，Workflow extension 会同步枚举并 hydrate agent directory 下所有持久化 Workflow，再在 hydrate 之后判断哪些 run 属于当前 Session。多数 run 对 child 是 foreign，但其文件读取和 JSON parse 成本已经发生。

## 证据
- `extensions/workflows/index.ts:992-998`：启动期恢复 persisted runs；
- `extensions/workflows/dashboard.ts:87-127,142-156`：同步读取、parse 和 normalize workflow artifacts；
- ownership/session 过滤晚于 hydration。

成本近似：`O(child starts × 全部 persisted workflow bytes)`。历史 run 越多，Direct/Workflow child 启动越慢，并同步阻塞事件循环。

## 期望
- 在读取完整 artifact 前使用可信、轻量、可验证的 owner/session index 过滤；
- child-safe capability binding 不恢复与该 child 无关的 operator history；
- index 缺失或损坏时 fail closed，不把不可信 metadata 当 ownership；
- benchmark：0/100/1,000 个 foreign runs 下启动 1/4/8 children，测同步阻塞、读取 bytes、parse 次数和 startup latency；
- restore/replay/dashboard 的 canonical artifacts 语义保持不变。

Sun-GLiang:
## Summary

- derive parent-only OpenPI extensions from the existing tool surface, owner source paths, and child exclusion contract
- filter only verified package-local parent-only extensions before child lifecycle binding
- preserve child-safe OpenPI tools, third-party extensions, dynamic structured output tools, and existing parent Workflow behavior
- add deterministic regression coverage and a workflow child-startup benchmark

## Validation

- bun run check
- bun run test (875 Node tests and 30 Vitest tests)
- bun run benchmark:workflow-child-startup (all 9 scenarios: 0 workflow reads, 0 bytes, 0 parses)
- Pi 0.84.1 runtime smoke for Direct and Workflow children
- parent Dashboard smoke confirms foreign Workflow history is not restored into a new request

Closes #183


agnitum2009 (dismissed):
Verified locally on the PR head (0f0f8db) — this is a clean, well-scoped fix for #183. Approving.

**Design**
- Deriving the parent-only extension set from the existing single source of truth (`OPENPI_TOOL_SURFACE` + `CHILD_EXCLUDED_TOOL_NAMES`) is the right move: a future owner whose entire surface is child-excluded is filtered automatically, and a *mixed* owner (any child-safe tool) is never filtered, so the existing denylist drift guard stays the fail-closed backstop.
- Path matching degrades safely: if an owner source path can't be canonicalized, the extension still loads and its tools are still stripped by `childToolPolicy` — fail-open on the perf win, fail-closed on correctness.
- `realpathSync` on both sides handles symlinked installs; `extensionsOverride` is a real `LoadExtensionOptions` hook in the pinned SDK, and filtering before child lifecycle binding is the correct seam.

**Verified by running it**
- `bun run typecheck` (only the two pre-existing file-search Effect warnings, unrelated) and `bun run lint` pass.
- Full suite: 875/875 Node tests + 30/30 Vitest, matching the PR's claim.
- `bun run benchmark:workflow-child-startup`: all 9 scenarios report 0 reads / 0 bytes / 0 parses; 8-child startup is flat across 0/100/1000 foreign runs (~545ms vs ~515ms), confirming the perf goal.
- New tests cover the important edges: third-party extensions survive tool-name collisions, child-safe extensions (fd, git_show) stay loaded, and dynamically registered `structured_output` tools survive binding.

**Notes (non-blocking)**
- The regression test monkeypatches global `fs.readFileSync`/`JSON.parse` with `syncBuiltinESMExports`; it restores state in `finally` correctly, but any future parallel test runner would need care. Acceptable for a targeted regression test.
- The benchmark script is opt-in via a separate npm script and does not run in CI; the deterministic fs/JSON instrumentation in the unit test is what guards the invariant over time. Good split.

Sun-GLiang:
Hi maintainers — this PR is ready for another review.

It addresses #183 by skipping verified parent-only workflow extensions during child-session binding while preserving child-safe OpenPI tools, third-party extensions, and dynamically registered tools. The full validation suite and runtime smoke tests pass, and the benchmark confirms that child startup no longer reads or parses unrelated workflow history.

When you have time, I’d appreciate a maintainer review, particularly of the extension classification and child-session authority boundary. Thank you!

tt-a1i (approved):
Re-reviewed exact head 053265a4a9faede21fdaa79f3abb7deb23753a0d after updating onto main ba26d325b4e7537fe38a2a426354cbeef9774a9e. The optimization derives parent-only extension owners from the existing tool-surface classification, preserves child-safe and third-party extensions, and fails closed when package/path identity is unverifiable. Local validation passed: focused 25/25, bun run check, 1060 Node tests passed with 1 platform skip, and 30/30 Vitest tests. The startup benchmark reported zero Workflow artifact reads, bytes, and JSON parses for all 1/4/8-child scenarios at 0/100/1000 foreign retained runs, so the performance win does not depend on weakening child tool policy. Fresh required CI passed on Node 22.19.0, Node 24, and Windows. No blocking findings.

PR #236, opened 2026-08-28
fix(background-terminals): sanitize model-facing output

627150795:
## Problem

The model-facing `bg_status` and background-terminal completion messages passed retained stdout/stderr directly to `truncateTail`. The TUI and watch paths already sanitize terminal text, but these prompt paths could expose ANSI/OSC, bidirectional, and other control characters. Truncating first could also leave an unmatched control sequence fragment in the model-visible tail.

## Value

Keep model-facing terminal output readable and safe by removing terminal control text before applying the existing output-size and line limits.

## Approach

- Reuse the existing shared `sanitizeTerminalText` helper.
- Sanitize once in the shared output-section path used by both `bg_status` and completion follow-ups.
- Keep raw captured output and spill files unchanged for evidence and `/ps` viewing.
- Add regression coverage for ANSI, OSC, bidi, and control characters, including a payload larger than the display cap to prove sanitization happens before truncation.

Related to #72

## Validation

- `node --test --experimental-strip-types tests/extensions/background-terminals/prompt.test.ts` — 7/7 passed
- `bun run lint` — passed
- `bun run typecheck` — passed; existing file-search Effect warnings remain outside this change
- `git diff upstream/main...HEAD --check` — passed

## Impact

Only model-facing stdout/stderr previews change. Raw logs, spill files, TUI rendering, process lifecycle, session disk budgeting, and Windows termination behavior are unchanged.

### DIFF
--- a/extensions/background-terminals/src/prompt.ts
+++ b/extensions/background-terminals/src/prompt.ts
@@ -13,6 +13,7 @@ import {
   type TerminalSnapshot,
 } from "./domain.ts";
 import { MAX_RUNNING, type KillResult } from "./manager.ts";
+import { sanitizeTerminalText } from "../../shared/terminal-text.ts";
 
 /** bg_status stdout tail. */
 export const STATUS_STDOUT_MAX = 16 * 1024;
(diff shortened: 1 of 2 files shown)


627150795:
Closing this PR as superseded by the earlier canonical fix in PR #149, which was merged into main and covers the same model-facing terminal-output sanitization (plus the remaining Issue #72 scope). The current upstream implementation uses the retained model-safe output projection from #149. This PR is therefore a duplicate; the claim and focused test work remain in the public history.

Issue #112, opened 2026-08-24
workflows: replay path returns ok regardless of acceptance verdict (defense-in-depth gap)


tt-a1i:
## Description

Verified by read-only audit (`seal/gpt-5.6-sol` subagent, verdict: PARTIAL — branch asymmetry real, ordinary resume not affected).

**Live path**: `index.ts:1847-1866` calls `applyAcceptance`; any `rejected`/`missing`/`malformed` ledger makes `outcomeOk=false`, the invocation outcome `error`, and record state `error` (`acceptance.ts:278-299` converts every non-accepted verdict into failure). Live persistence writes the result artifact and journal entry **only when** `outcomeOk` (`index.ts:1867-1877`, `index.ts:1910-1925`).

**Replay path**: `index.ts:1634-1685` evaluates acceptance for display but unconditionally records `state="done"`, registers `ok:true`, and returns `ok:true`.

## Reachability

Ordinary resume **cannot** reach the bad replay behavior:

- A rejected/missing/malformed live outcome never enters the journal (failures are rerun, not journaled).
- The acceptance contract and effective schema participate in the replay key (`index.ts:1610-1630`), preventing reuse under a changed contract.

The gap is defense-in-depth: `parseJournal` (`journal.ts:29-34, 172-188`) does not validate acceptance, so a manually modified/corrupted journal — or any future producer violating the success-only invariant — would be replayed as `ok:true`/`done` regardless of verdict.

## Suggested fix

Run the cached payload through the same `applyAcceptance` path as a live outcome: only mark the record done and return `ok:true` when the verdict is `accepted`; otherwise persist ledger/error/failed agent state consistently with normal execution. Alternatively/additionally, validate acceptance at journal parse time.

## Test coverage gap

`replay-safety.test.ts` has no acceptance coverage; `execute.e2e.test.ts:473-555` covers only successful acceptance-free replay.

Found by automated workflow audit (run `wf_56028a5b018b`).

627150795:
Hi! I am taking this issue next. I will make the replay path apply the same acceptance verdict as live execution, so rejected/missing/malformed acceptance cannot be registered as `ok: true`/`done`; I will add focused replay coverage while preserving replay identity, journaling, artifact persistence, and delivery semantics. The repository does not allow external contributors to self-assign this issue, so this comment is the claim record.

627150795:
Implementation is ready for review in PR #241: https://github.com/tt-a1i/openpi/pull/241

[...]

PR #225, opened 2026-08-28
fix(workflows): reject failed acceptance during replay

REI-330:
## Problem

Workflow replay evaluated cached acceptance data for display but still recorded the agent as `done` and returned `ok: true` when the verdict was rejected, missing, or malformed. A corrupted or manually modified journal could therefore bypass the live path's success-only invariant.

On Windows, persisted agent-result references also used host path separators even though workflow references are portable logical paths.

Fixes #112

## Value

Replay now enforces the same acceptance contract as live execution. Invalid cached results cannot become successful agent records or produce successful artifacts, references, or journal entries. Newly persisted artifact references are stable across platforms.

## Approach

- Run cached structured output through the shared `applyAcceptance` path.
- On acceptance failure, retain the acceptance ledger and error, transition the invocation to rejected/settled/error, mark the agent record as `error`, and return `ok: false` before success persistence.
- Keep artifact persistence, result-reference registration, journal append, and `replayed` state exclusive to accepted results.
- Persist artifact references with `/` while continuing to use host-native joins for filesystem writes.
- Add an end-to-end regression that tampers a successful journal entry into a rejected acceptance result and verifies the complete fail-closed state.

## Validation

- `node --test --experimental-strip-types tests/extensions/workflows/acceptance.test.ts tests/extensions/workflows/artifacts.test.ts tests/extensions/workflows/execute.e2e.test.ts` - 21/21 passed.
- `bun x biome format extensions/workflows/artifacts.ts extensions/workflows/index.ts tests/extensions/workflows/execute.e2e.test.ts` - passed, no fixes required.
- `bun x biome lint extensions/workflows/artifacts.ts extensions/workflows/index.ts tests/extensions/workflows/execute.e2e.test.ts --error-on-warnings` - passed.
- `bun run typecheck` - passed; the 12 existing Effect diagnostics in `extensions/file-search/src/binaries.ts` remain warnings.
- `git diff --check origin/main...HEAD` - passed.
- `bun run check` - not green on this Windows checkout because repository-wide format checking reports CRLF differences in approximately 252 untouched files.
- `bun run test` - 893/933 passed, 36 failed, 4 skipped. Failures are outside the changed Workflow paths and are concentrated in Windows shell, path-separator, temporary-directory, and Git-environment assumptions; the focused Workflow suite above passes.
- Manual Pi/TUI smoke was not run; this change is covered through deterministic persistence and replay tests.

## Impact

- User-visible behavior: replayed agent results with failed acceptance are shown as errors instead of successes.
[...]

PR #241, opened 2026-08-28
fix(workflows): honor acceptance verdicts during replay

627150795:
## Problem

The replay path evaluated an acceptance contract only for display, then marked a cached agent result as done and returned `ok: true` even when the replayed acceptance verdict was rejected, missing, or malformed. Related to #112.

## Value

Corrupted or manually edited journals now fail closed instead of silently presenting an unaccepted replay as a successful agent result.

## Approach

- Route replayed cached results through the same `applyAcceptance` decision used by live execution.
- Persist the acceptance ledger and return an error state without creating an artifact, handoff reference, or replay journal entry when the verdict is not accepted.
- Keep successful replay behavior unchanged, including artifact persistence and re-journaling.
- Use stable run-relative forward-slash artifact references so replay handoffs remain valid on Windows.
- Add end-to-end coverage for accepted, rejected, missing, and malformed replay verdicts.

## Validation

- `node --test --experimental-strip-types tests/extensions/workflows/execute.e2e.test.ts` — 12 passed.
- `node --test --experimental-strip-types tests/extensions/workflows/acceptance.test.ts` — passed.
- `bun run lint` — passed.
- `bun run typecheck` — passed with the repository's existing warnings in `extensions/file-search`.
- `git diff --check` — passed.
- The workflow artifact test retains an existing Windows-only expectation mismatch: directory replacement reports `EPERM` rather than `EISDIR`.

## Impact

- User-visible behavior: replayed non-accepted results are reported as failures instead of successes.
- Runtime/lifecycle: rejected replays end in the error state and do not register successful handoffs.
- Persisted data: accepted replay artifacts and journal chaining are preserved; rejected replays do not create success artifacts.
- Compatibility/risk: no acceptance contract or live-execution semantics changed; malformed journal payloads now fail closed.

### DIFF
--- a/extensions/workflows/artifacts.ts
+++ b/extensions/workflows/artifacts.ts
@@ -127,10 +127,7 @@ export function persistWorkflowAgentResult(
   index: number,
   result: { output: string; structured?: unknown },
 ) {
-  const artifact = path.join(
-    "agent-results",
-    `agent-${String(index).padStart(4, "0")}.json`,
-  );
+  const artifact = `agent-results/agent-${String(index).padStart(4, "0")}.json`;
   const encoded = encodeCompleteJson(
     {
       output: result.output,
--- a/extensions/workflows/index.ts
+++ b/extensions/workflows/index.ts
@@ -78,7 +78,6 @@ import {
   acceptanceInstruction,
   acceptanceSchema,
   applyAcceptance,
-  evaluateAcceptance,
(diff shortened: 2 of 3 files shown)


627150795:
Closing this duplicate PR: #225 predates it and covers the same Issue #112 replay-acceptance fix. Please review the earlier contribution instead. The branch is left intact for local reference.

Issue #213, opened 2026-08-28
[Proposal] Mechanize the configuration contract: five-way sync check, provenance script, and a discipline ledger


liwen-aaa:
<html>
<body>
[...]

PR #242, opened 2026-08-28
feat(contract): mechanize config provenance checks

627150795:
## Problem\n\nThe package configuration and runtime-provenance contracts were documented but not mechanically checked, so a future configuration change could drift between defaults, the setup writer, status output, and the two user-facing documents.\n\n## Value\n\nMake the existing contributor contract fail fast during the normal check round, and give contributors one reproducible command to prove which checkout and Pi package source they are diagnosing.\n\n## Approach\n\n- Add a configuration contract checker using the existing Acorn dependency, including focused fixture self-tests.\n- Cover the 14 persisted configuration fields, including the optional suggestion model, across defaults/type, setup writer, status projection, README, and SETUP markers.\n- Add a provenance command that compares the checkout revision and the OpenPI source reported by pi list.\n- Add an append-only engineering-discipline ledger with validation that referenced checks exist and are wired into CI.\n- Preserve runtime behavior and keep all changes in development checks and documentation.\n\n## Validation\n\n- node scripts/check-config-contract.mjs — passed (14 persisted fields)\n- node scripts/check-config-contract.mjs --self-test — passed (8 fixture checks)\n- node scripts/check-discipline-ledger.mjs — passed (12 ledger rows)\n- node scripts/provenance.mjs — passed (checkout and pi list source match)\n- node --test --experimental-strip-types tests/scripts/check-config-contract.test.ts — passed\n- bun run lint — passed\n- bun run typecheck — passed\n- git diff --check — passed\n- Full Windows bun run check/test runs remain affected by existing CRLF, path, symlink, and process-lifecycle baseline failures; no runtime files were changed here.\n\n## Impact\n\n- User-visible behavior: None.\n- Model-visible context/tools: None.\n- Runtime/lifecycle: None.\n- Persisted config/data: None.\n- Compatibility/risk: Development-time checks only; the existing acorn dependency is reused and no new dependency is added.\n\nFixes #213


627150795:
补充代码审计结果：已修复可选配置字段漏检和 append-only 历史行可被重编号绕过的问题；新增回归测试，配置检查、ledger 检查、lint/typecheck 全部通过。PR 仍等待维护者审核。

liwen-aaa:
Nice to see this moving — thanks for picking it up, and the three-check structure matches the proposal well.

   One review point on test depth before this gets too far, since a contract   checker that cannot fail is worse than none:

   The current config test (`tests/scripts/check-config-contract.test.ts`) shells   out to `--self-test` and asserts exit code 0. That proves the script can run   its own fixtures, but it does not prove the checker actually catches drift on  the real contract shape — a regression that makes the checker pass-everything  would stay green.

   I built a proof-of-concept of this same check earlier (referenced in the  issue) and the cases that turned out to matter are the ones where the checker  must go red:

   1. **Layer-1 miss**: a leaf field defined in `setup-config.ts` but absent from
      `setup/index.ts` (the typed write tool) — the "new field never wired" shape.
   2. **Doc zero-hit**: a field whose registered doc keywords match nothing in
      SETUP.md / README.md (doc wording drifted or the field was never documented).
   3. **Unregistered field**: a field added without registering its doc keywords —
      this catches the mapping itself going stale.
   4. **Entry removed**: README losing the `/openpi-setup` section entirely.

   Each of these is cheap to fixture (small inline fake of the three files) and each is the shape that actually bites contributors later. The current  `--self-test` covers none of them — it verifies the fixture the script itself owns.

   Happy to share the fixture shapes I validated (6 cases, all green on the  current tree, all red on the injected fault) if useful — or to review the checker logic itself.

   The discipline-ledger test (delete + renumber rejection) is solid, by the way.
 ```

627150795:
行为复核补充（2026-08-29）：根据审查意见补强配置契约检查的反例自测。现在自检明确验证以下漂移会失败：写入层缺字段、SETUP.md 文档关键词失配、默认配置出现未登记字段、README 配置区段整体缺失；原有可选字段、未知 marker、历史 ledger 删除/重编号用例仍保留。

本地验证：
- `node scripts/check-config-contract.mjs --self-test`：12 组 fixture 通过
- `node --test --experimental-strip-types tests/scripts/check-config-contract.test.ts`：1/1 通过
- `node scripts/check-config-contract.mjs`：14 persisted fields 通过
- `node scripts/check-discipline-ledger.mjs`：12 append-only rows 通过
- `bun run lint`：通过
- `bun run typecheck`：通过
- `git diff --check`：通过

新增提交：`5069c3c`。这次修的是“检查器测试是否真的能抓住漂移”，不改变 OpenPI 运行时行为。

liwen-aaa:
The ledger hardening in `5069c3cf` is the right shape — a static `HISTORICAL_LEDGER` baseline catches deletion/renumbering/rewrite, and `check-discipline-ledger.test.ts` proves the red case instead of just asserting the script runs. That is exactly the "make manual judgement minimal" principle applied correctly.

   The same principle should now be applied to the config checker, because the two halves of this PR are now inconsistent:

   - The ledger: the ground truth is the historical baseline, drift is detected by comparison, nothing is hand-maintained beyond the baseline itself.
   - The config checker: the ground truth is a 14-row hand-registered table (`CONFIG_FIELD_CONTRACT`), where each row carries five token groups (writerTokens / statusTokens / readmeTerms / setupTerms / configTokens) that encode "this doc wording means this field" — judgement no checker can verify mechanically.

   Concrete failure: add a new persisted field and you must edit code + the table + doc markers in lockstep. Forget the table and the checker is silently blind to that field — it never learns the field exists, so a missing README entry passes. The ledger fix you just made is a regression test against exactly this class of failure; the config table is still vulnerable to it.

   Most of the table can be derived. `extractDefaultFieldPaths` already correctly derives leaf fields from `DEFAULT_SETUP_CONFIG` via Acorn — that is the right pattern. writerTokens / statusTokens can likewise be derived by scanning `extensions/setup/index.ts` and the status formatter for each derived path instead of hand-registering tokens. What genuinely cannot be derived — doc wording ↔ field mapping — is the smallest part and could be a documented, review-required list rather than a silent code table.

   One more fragility worth addressing while the file is open: `extractOptionalConfigPaths` hand-rolls regex parsing of the `interface MyPiSetupConfig` shape (the "ponytail" comment acknowledges it). Any reformat — multi-line type, a comment containing a brace, nested object — silently yields wrong paths, and a wrong checker is worse than none because it stays green while real drift passes. Since Acorn already handles expressions, the optional fields could move into `DEFAULT_SETUP_CONFIG` with a sentinel so the existing extractor covers them, or the interface parse could be replaced.

   Net: the drift-detection direction and the new ledger test are good; the config table and the regex parser are the two places a future contributor will break, and both are cheap to fix now while the shape is still fresh.

627150795:
再次复核：新增提交 `0b4c25b` 已通过远端 CI（Node 22.19.0、Node 24、Windows Background terminals 全部通过）。之前的失败确认为新增脚本的换行/格式问题，已修正；配置检查反例自测仍为 12 组通过，未改变 OpenPI 运行时行为。

627150795:
代码审计补丁已追加到本 PR：

- 复现并修复了配置检查器的误报：仅在真正的 `buildConfig` 写入区验证字段，参数 schema 中残留 `params.xxx` 不再被当成已落盘。
- 新增反例：保留 schema 参数、删除实际 writer 赋值时必须失败。
- 远程提交：`daa33f3`。
- 验证：配置自测 13 项通过，真实配置 14 项通过，Node 定向测试 1/1、lint、typecheck 均通过；当前远程 Node 22、Node 24、Windows CI 全部通过。

这只修正契约检查的有效性，不改变 OpenPI 运行时行为。

somewan820 (approved):
已完成静态审查。config-contract 与 discipline ledger 检查器覆盖默认配置、typed setup writer、status projection、README/SETUP marker、append-only ledger 及 CI wiring；新增脚本均有测试或自检入口，未发现 P0-P2 或无用实现。P3：静态 token/轻量 parser 对未来源码排版变化较敏感，后续可考虑更强的结构化 fixture。未运行测试。

Issue #193, opened 2026-08-25
docs(session): define lifecycle, recovery, replay, and side-effect boundaries


tt-a1i:
# docs(session): define lifecycle, recovery, replay, and side-effect boundaries

**Proposed issue type:** documentation / architecture contract  
**Recommended artifact:** `docs/design/SESSION_LIFECYCLE_AND_RECOVERY_CONTRACT.md`  
**Implementation impact:** none required  
**Audit coverage:** five source workstreams completed, covering seven grouped systems; failed workstreams: **[]**

## 1. Executive conclusion

OpenPI does not need another Session store, checkpoint engine, rewind mechanism, compaction implementation, or provider continuation layer. It needs one cross-cutting lifecycle contract that distinguishes:

1. Pi’s canonical Session transcript;
2. active runtime ownership;
3. OpenPI Workflow execution artifacts;
4. provider-private continuation/cache state;
5. filesystem and Git state;
6. model-visible context projections;
7. operator-facing UI projections.

The core invariant is:

> OpenPI extends Pi at its native lifecycle seams. Pi remains the sole owner of Session identity, transcript persistence, branch selection, resume, fork, and compaction. OpenPI may persist package-owned execution artifacts and branch-scoped custom entries, but must not create a second Session store or infer execution completion from labels, dashboards, process disappearance, or provider state.

A second essential invariant is:

> Workflow replay means persisted-result reuse for a previously proven read-only child invocation. It is never Session resume, child conversation reconstruction, operator-memory restoration, provider cache reuse, or tool-call re-execution.

---

### Fixed-source facts

| System | Bounded revision or version | Verification status |
|---|---|---|
| OpenPI | `2a69d3f32994da4123f1312b7fa84ef3d6119be1` | Audit-designated checkout revision |
| Pi | `@earendil-works/pi-coding-agent@0.84.3` | Installed package inspected |
| Codex | `a9e447a69dee4f2789dd8d8c776e314772c1f049` | Git object independently verified and inspected |
| Gemini CLI | `812f7a2bcf20b6e80e2e50c3c8fa8e26567bc1e8` | Git object independently verified and inspected |
| OpenCode | `8615731d46153dd29b89e205fb55b2cc16205cb0` | Git object independently verified and inspected |
| Hermes | `cbd8de8ad64530be01efea23b7764d5c37c634ed` | Git object independently verified and inspected |
[...]

PR #244, opened 2026-08-28
docs(session): define lifecycle and recovery contract

627150795:
## Problem\n\nSession, Workflow, Replay, provider, filesystem/Git, and UI state were documented in separate places, making it easy to confuse a transcript branch with runtime ownership, a result artifact with Session history, or a process exit with successful completion.\n\n## Value\n\nProvide one reviewable lifecycle and recovery contract that preserves Pi as the canonical Session owner, makes uncertainty explicit, and gives contributors a shared boundary for recovery, replay, retention, and side effects.\n\n## Approach\n\n- Add the requested cross-plane contract under docs/design with evidence labels and bounded revision limitations.\n- Separate Pi Session identity, model context, OpenPI Workflow execution, provider state, filesystem/Git state, UI projection, and delivery evidence.\n- Define resume, live rejoin, tree, fork, clone, compaction, context pivot, child settlement, Workflow recovery, delivery, and Replay independently.\n- Document parent/child fail-closed authority, terminal evidence hierarchy, uncertain outcomes, retention/deletion boundaries, and explicit non-goals.\n- Link the contract from README, the Workflow graph design, and Subagent/Workflow guidance without adding runtime code or a second persistence/control plane.\n\n## Validation\n\n- git diff --check — passed\n- bun run lint — passed\n- bun run typecheck — passed\n- No runtime behavior was changed; this is a documentation/design-only PR.\n\n## Impact\n\n- User-visible behavior: None.\n- Model-visible context/tools: None.\n- Runtime/lifecycle: None.\n- Persisted config/data: None.\n- Compatibility/risk: Documentation only; the contract explicitly preserves existing Pi/OpenPI ownership and defers implementation until a concrete invariant requires it.\n\nFixes #193

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,6 +314,9 @@ Workflow 在清理隔离 checkout 前原子保存有界 Handoff Manifest：track
 
 设计细节见 [`docs/design/WORKFLOW_INVOCATION_GRAPH.md`](docs/design/WORKFLOW_INVOCATION_GRAPH.md)。
 
+Session、Workflow、Replay、恢复与文件副作用的跨平面边界见
+[`docs/design/SESSION_LIFECYCLE_AND_RECOVERY_CONTRACT.md`](docs/design/SESSION_LIFECYCLE_AND_RECOVERY_CONTRACT.md)。
+
 ---
 
 ## 连续工作，而不是堆 Context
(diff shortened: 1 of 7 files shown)


tt-a1i:

PR #257, opened 2026-08-28
docs(control): define Pi-native safe pause contract

627150795:
## Problem
OpenPI has separate cancellation controls, but no contract for a safe pause that stops new provider/tool work across Pi-owned execution paths without pretending that in-flight work or external terminals were frozen.

## Value
This gives maintainers and operators precise pause states, receipts, safe boundaries, scope ownership, and a clear distinction from cancel/stop. It prevents an extension-level approximation from making a false global-safety claim.

## Approach
Add a focused design contract for Pi-owned gate ownership, requested/partially-parked/fully-parked/resuming states, provider/tool/spawn/completion boundaries, Background Terminal independence, lifecycle edge cases, and the implementation/test gate. No runtime pause flag is added.

## Validation
- git diff --check
- bun run lint
- bun run typecheck
- Runtime pause tests not run because this PR is design-only and records the required Pi seam first.

## Impact
- Runtime/lifecycle: no change.
- Compatibility/risk: docs-only; current per-owner controls remain unchanged.

Addresses #161

### DIFF
--- a/docs/design/PI_NATIVE_SAFE_PAUSE_CONTRACT.md
+++ b/docs/design/PI_NATIVE_SAFE_PAUSE_CONTRACT.md
@@ -1,0 +1,140 @@
+# Pi-native safe-pause contract
+
+Status: proposed research/design for [Issue #161](https://github.com/tt-a1i/openpi/issues/161). This document does not add a pause implementation.
+
+## Problem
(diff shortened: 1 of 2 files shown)


tt-a1i:

PR #259, opened 2026-08-29
docs(capabilities): define on-demand lsp contract

627150795:
## Problem

OpenPI optional capabilities are loaded explicitly and remain stable for the Session, but the repository currently has no LSP client or server contract. An LSP integration could otherwise start processes implicitly, widen file access, or expose an unbounded protocol surface.

## Value

Define a Pi-native boundary for a future on-demand LSP capability: explicit activation, session/workspace ownership, read/write separation, cancellation and cleanup, bounded results, and fail-closed behavior.

## Approach

- Add docs/design/ON_DEMAND_LSP_CAPABILITY_CONTRACT.md.
- Keep LSP loading separate from server startup and separate read queries from future write operations.
- Specify path containment, trust/configuration, timeout, cancellation, crash, cleanup, and model-visible evidence rules.
- Mark the document as a design-only first pass; no LSP runtime, dependency, or current capability list is changed.

## Validation

- git diff --check
- bun run lint
- bun run typecheck

## Impact

Documentation only. Text search and read-only git inspection remain the supported search capabilities. This is a focused first pass for #163 and does not claim an LSP runtime is implemented.

Addresses #163

### DIFF
--- a/docs/design/ON_DEMAND_LSP_CAPABILITY_CONTRACT.md
+++ b/docs/design/ON_DEMAND_LSP_CAPABILITY_CONTRACT.md
@@ -1,0 +1,159 @@
+# On-Demand LSP Capability Contract
+
+Status: design-only first pass for [issue #163](https://github.com/tt-a1i/openpi/issues/163). This document defines a future capability boundary; the current OpenPI runtime has no LSP client, server, or LSP dependency.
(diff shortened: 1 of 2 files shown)


tt-a1i:

PR #260, opened 2026-08-29
docs(subagents): define structured result contract

627150795:
## Problem

Direct Subagents and Workflow children currently expose several representations of one run: the native child transcript, the normalized snapshot, the bounded parent projection, and optional workflow artifacts. Without an explicit boundary, a projection or JSON-shaped payload can be mistaken for the canonical result or for completion evidence.

## Value

Define a small, versioned vocabulary for future structured child results while preserving the current text-only runtime behavior. This gives direct Subagents and Workflows a shared basis for partial outcomes, validation, evidence, and bounded delivery.

## Approach

- Add docs/design/STRUCTURED_SUBAGENT_RESULT_CONTRACT.md.
- Define canonical result ownership, parent projections, optional schema validation, failure/partial-result semantics, evidence boundaries, and budget/privacy rules.
- Mark the document as a design-only first pass; no runtime API, dependency, or result schema is changed.

## Validation

- git diff --check
- bun run lint
- bun run typecheck

## Impact

Documentation only. The current implementation remains the source of truth for shipped behavior. This is a focused first pass for #155 and does not claim the complete runtime feature is implemented.

Addresses #155


tt-a1i:

PR #264, opened 2026-08-29
docs(debugging): define dap failure benchmark

627150795:
## Problem

Issue #168 needs evidence about whether on-demand DAP improves diagnosis over logs and tests, but a prompt-only approximation or a debugger self-report would not be a valid comparison.

## Value

Define a failure-class benchmark that measures diagnosis quality, time, tokens, false leads, debugger failures, cleanup, and side effects.

## Approach

- Add docs/design/DAP_FAILURE_CLASS_BENCHMARK_PROTOCOL.md.
- Define logs and tests, on-demand DAP, and combined arms with a fixed failure corpus.
- Freeze runtime, model, tool, permission, verifier, and timeout variables.
- Record diagnosis rubric, DAP safety boundaries, overhead, failures, and a decision gate.
- Keep the first step protocol-only; mark DAP not-run when unavailable.

## Validation

- git diff --check
- bun run lint
- bun run typecheck

## Impact

Documentation only. This is a focused research slice for #168 and contains no DAP runtime or experimental result.

Addresses #168


tt-a1i:

PR #265, opened 2026-08-29
docs(runtime): define package-wide admission leases

627150795:
## Problem

Direct Subagents, Workflow runs, and Background Terminals enforce local limits independently, but OpenPI has no shared fact for package-wide model, process, worktree, or artifact usage.

## Value

Define a shared admission lease boundary without merging owner managers or introducing a global scheduler.

## Approach

- Add docs/design/PACKAGE_WIDE_ADMISSION_LEASE_CONTRACT.md.
- Define lease identity, atomic acquire/reject, release, cancellation, owner loss, configuration shrink, fairness, and bounded diagnostics.
- Preserve each owner lifecycle, local limits, cancellation policy, artifacts, and user-facing result.
- Keep the first step design-only; do not add a runtime governor or change current limits.

## Validation

- git diff --check
- bun run lint
- bun run typecheck

## Impact

Documentation only. This is a focused design slice for #159 and does not implement a package-wide scheduler.

Addresses #159

### DIFF
--- a/docs/design/PACKAGE_WIDE_ADMISSION_LEASE_CONTRACT.md
+++ b/docs/design/PACKAGE_WIDE_ADMISSION_LEASE_CONTRACT.md
@@ -1,0 +1,186 @@
+# Package-Wide Admission Lease Contract
+
+Status: design-only first pass for [issue #159](https://github.com/tt-a1i/openpi/issues/159). This document defines a shared resource-admission boundary; it does not merge the existing managers, add a scheduler, or change current limits.
+
+## Problem
+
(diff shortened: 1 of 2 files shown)


tt-a1i:
Thank you for preserving this research. We are closing this PR as a source record for the Capability, authority, and resources consolidation tracked in #277. This is not a finding that the content is wrong. The current queue contains overlapping contracts from one stale base, so we will extract only settled boundaries into at most one successor record, citing this PR explicitly. Please do not spend time rebasing this branch.

PR #266, opened 2026-08-29
docs(editing): define structural query edit boundary

627150795:
## Problem

Text search and ordinary editing cannot reliably distinguish syntax from comments or strings, while structural rewrites would introduce a new multi-file mutation authority.

## Value

Define separate read-only structural query and preview-first mutation contracts with stale-file fail-closed behavior.

## Approach

- Add docs/design/STRUCTURAL_QUERY_PREVIEW_EDIT_CONTRACT.md.
- Separate query, preview, and future apply capabilities.
- Define proposal identity, parser and limit partiality, path and child authority, multi-file recovery, and hash verification.
- Keep the first step design-only; do not add an AST parser, dependency, or runtime mutation.

## Validation

- git diff --check
- bun run lint
- bun run typecheck

## Impact

Documentation only. This is a focused design slice for #165 and does not implement an AST query or edit runtime.

Addresses #165


tt-a1i:

PR #268, opened 2026-08-29
docs(compute): define persistent eval benchmark

627150795:
## Problem

Issue #164 needs evidence about persistent computation and Tool Bridge value before introducing a new code-execution authority.

## Value

Define a paired benchmark that separates persistent state, structured output, and Tool Bridge while measuring quality, cost, lifecycle, timeout, reset, cleanup, and permission risk.

## Approach

- Add docs/design/PERSISTENT_EVAL_BENCHMARK_PROTOCOL.md.
- Compare one-shot tools, persistent compute without a bridge, and a restricted read-only bridge.
- Freeze task, model, runtime, permission, and verifier variables and record state, output, latency, and cleanup evidence.
- Keep the first step protocol-only; mark unavailable arms as not-run.

## Validation

- git diff --check
- bun run lint
- bun run typecheck

## Impact

Documentation only. This focused research slice for #164 does not add a persistent Kernel, arbitrary eval, or Tool Bridge.

Addresses #164


tt-a1i:

PR #270, opened 2026-08-29
feat(ci): notify Feishu for new pull requests

junwu168:
## Problem

New pull requests were not announced automatically in the project Feishu group, and public PR metadata could be interpreted as Feishu text markup if forwarded verbatim.

## Value

The repository can notify the Feishu group immediately while keeping the webhook and signing secret protected, while hostile titles or other metadata cannot inject mentions or extra message fields.

## Approach

Add a `pull_request_target` workflow for opened and ready-for-review events. It checks out only the trusted `github.workflow_sha` with credentials disabled, never PR code, then signs and sends a bounded six-line plain-text notification. All public PR metadata is projected through a Unicode-aware sanitizer before the HTTP payload is built. The workflow accepts both current and legacy Feishu success response formats.

## Validation

- Focused formatter and real local HTTP payload tests: 7/7 passed
- `bun run check`: passed (258 files)
- `bun run test`: passed (966 passed, 1 skipped)
- `git diff --check`: passed
- Two independent final reviews: no findings
- Adversarial entity and Unicode matrix: 324 combinations, 0 bypass
- Live Feishu notification: not sent; runtime acceptance requires the workflow on the default branch and a later matching PR event

## Impact

User-visible behavior: Feishu receives one message for newly opened non-draft PRs and another whenever a draft is marked ready. Repeated draft-to-ready transitions intentionally produce repeated notifications.

Model-visible context/tools: None.

Runtime/lifecycle: Adds one bounded GitHub Actions job per relevant PR event.

Persisted config/data: Requires `FEISHU_PR_BOT_WEBHOOK` and `FEISHU_PR_BOT_SECRET` repository secrets.

Compatibility or risk: Benign notification text, signing, response handling, and missing/partial-secret behavior are preserved. Hostile, control, bidi, and overlong metadata is safely bounded.

### DIFF
--- a/.github/actions/feishu-pr-notification/index.cjs
+++ b/.github/actions/feishu-pr-notification/index.cjs
@@ -1,0 +1,151 @@
+const crypto = require("node:crypto");
+const fs = require("node:fs");
+
+const FIELD_LIMIT = 256;
+const REVIEWER_LIMIT = 128;
+const REVIEWERS_LIMIT = 512;
+const URL_LIMIT = 512;
+const CONTROL_OR_LINE_SEPARATOR = /[\u0000-\u001f\u007f-\u009f\u2028\u2029]/u;
+const BIDI_CONTROL = /[\u061c\u200e\u200f\u202a-\u202e\u2066-\u2069]/u;
+
+function truncateCodePoints(value, maxCodePoints) {
+  const points = Array.from(value);
+  if (points.length <= maxCodePoints) return value;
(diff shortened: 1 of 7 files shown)


tt-a1i (approved):
Reviewed exact head 7e1718a: trusted workflow checkout, bounded Feishu metadata projection, real HTTP payload coverage, full check and test suite green; two independent final reviews reported no findings.

PR #318, opened 2026-08-31
perf(workflows): append replay journal incrementally

### LINKED ISSUE #184: perf(workflows): replay journal 反复完整序列化导致近二次开销
## 问题
Workflow replay journal 每追加一次成功/重放调用，都会重新序列化 journal suffix，并在超限时反复 slice/重算完整 JSON 大小。接近 2 MiB journal 上限、调用数达到配置上限时，累计成本近似二次增长。

## 证据
- `extensions/workflows/index.ts:1730,1974`：agent call 持续追加 journal 并触发 persistence；
- `extensions/workflows/journal.ts:148-163`：每次对候选 suffix 完整 stringify/size，并循环丢弃旧条目；
- `extensions/shared/setup-config.ts:130-133`：单 run 调用量可达 1,024。

## 期望
- append 的正常路径不重新编码完整 journal；
- byte accounting 增量维护，eviction 采用 cursor/deque 或等价有界结构；
- 最终文件仍完整、原子、可验证，不能用 lossy 估算破坏 replay；
- malformed/legacy journal 继续 fail closed；
- benchmark：128/512/1,024 个接近单项上限的记录，测总序列化 bytes、CPU、allocations 和 final flush；增长应接近线性。

testikun:
> *This was generated by AI during triage.*

## Problem

Closes #184.

Workflow replay journals currently re-serialize the complete pretty-printed journal while trimming old entries. Near the 2 MiB cap, repeated `slice(1)` and full JSON size recalculation make append/eviction cost grow close to quadratically for large runs.

## Value

Large read-only Workflows can retain hundreds of replayable agent results. Incremental accounting keeps journal admission and oldest-entry eviction proportional to the number of new entries, while preserving the exact replay artifact and its atomic write path.

## Approach

- Add a `WorkflowJournalAccumulator` with cached per-entry JSON fragments, UTF-8 byte accounting, and a head-cursor queue.
- Append encodes only the new entry; eviction subtracts known bytes and releases dropped payload references immediately.
- Keep `boundedJournal()` as a compatibility adapter for array callers.
- Let persistence write the accumulator's complete canonical JSON through the existing atomic file writer.
- Preserve the v2 artifact format, oldest-first retention, duplicate-key replay order, and malformed/legacy fail-closed loading.
- Add a reproducible `benchmark:workflow-journal` script for 128/512/1024 entries.

## Validation

- `bun run check` — passed (config contract, discipline, format, lint, typecheck).
- `bun run test` under Node 24 — 1,076 passed, 1 skipped; Vitest 30/30 passed.
- Focused journal/artifact tests — 26/26 passed.
- `node --experimental-strip-types --expose-gc benchmarks/workflow-journal.ts`:
  - 128 entries: append 7.909 ms, 2,034,146 bytes retained.
  - 512 entries: append 22.447 ms, 2,034,149 bytes retained.
  - 1,024 entries: append 43.575 ms, 2,034,179 bytes retained.
  - Final serialized bytes exactly matched incremental accounting in every run.
- Before this change, the equivalent 4 KiB fixture took approximately 1,123 ms at 1,024 entries; the incremental append path took approximately 4 ms.

## Impact

- Runtime/lifecycle: journal append and eviction are incremental; replay semantics are unchanged.
- Persisted data: same v2 `journal.json` schema and canonical pretty JSON; writes remain atomic.
- Compatibility/risk: array-based callers remain supported; oversized/malformed/legacy behavior remains fail closed.

### DIFF
--- a/benchmarks/workflow-journal.ts
+++ b/benchmarks/workflow-journal.ts
@@ -1,0 +1,88 @@
+import { performance } from "node:perf_hooks";
+import {
+  createJournalAccumulator,
+  JOURNAL_MAX_BYTES,
+} from "../extensions/workflows/journal.ts";
+
+function readOption(name: string) {
+  const index = process.argv.indexOf(name);
+  if (index >= 0) return process.argv[index + 1];
+  return process.argv
+    .find((argument) => argument.startsWith(`${name}=`))
+    ?.slice(name.length + 1);
+}
+
+function parseCounts(raw: string | undefined) {
+  const counts = (raw ?? "128,512,1024")
+    .split(",")
+    .map((value) => Number.parseInt(value, 10));
+  if (counts.some((value) => !Number.isSafeInteger(value) || value <= 0)) {
+    throw new Error("--counts must contain positive integers");
+  }
+  return counts;
+}
+
+function parsePositiveInteger(raw: string | undefined, fallback: number) {
+  if (raw === undefined) return fallback;
+  const value = Number.parseInt(raw, 10);
+  if (!Number.isSafeInteger(value) || value <= 0) {
+    throw new Error("--entry-bytes must be a positive integer");
+  }
+  return value;
+}
+
+function heapUsed() {
+  return process.memoryUsage().heapUsed;
+}
+
+function cpuMicros(previous: NodeJS.CpuUsage) {
+  const usage = process.cpuUsage(previous);
+  return usage.user + usage.system;
+}
+
+const entryBytes = parsePositiveInteger(readOption("--entry-bytes"), 64 * 1024);
+
+for (const count of parseCounts(readOption("--counts"))) {
+  globalThis.gc?.();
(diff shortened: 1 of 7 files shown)


tt-a1i (approved):
Reviewed exact head 1e481d0ade0b9a6bd70da87eef6a17ac7bda935b. No blocking findings. The incremental accumulator avoids re-encoding retained entries while preserving canonical v2 JSON, oldest-first retention, replay ordering, and atomic persistence. Independent validation: focused 26/26, bun run check, full Node 1076 passed with one platform skip, Vitest 30/30; 6000 differential append checks matched the reference algorithm, and a 1024-append probe confirmed each payload was encoded once. Current Node 22.19.0, Node 24, and Windows CI passed. Non-blocking follow-ups: retain the encode-count probe as a regression test and clarify that benchmark flushMs measures JSON assembly rather than disk persistence.

Issue #304, opened 2026-08-30
test: eliminate Windows background-terminal process-test flakes in the full suite


yxr-2025:
## Problem

`scripts/run-tests.mjs` discovers every Node test file and passes all of them to one invocation:

```text
node --test --experimental-strip-types <all .test.ts files>
```

Node executes test files concurrently by default. The `background-terminals` manager tests are different from ordinary unit tests: on Windows they create real process trees, invoke and await `taskkill`, observe settlement callbacks, poll descendant liveness, and clean up OS resources. When other test files are creating or terminating processes at the same time, Windows scheduling and process-tree cleanup become timing-sensitive.

This produces intermittent failures around the same 3.3-3.8 second teardown window even though the affected tests are stable in isolation.

## Reproduction evidence

Observed on Windows with Node v24.12.0 and Bun 1.4.0:

- Each affected test passes when run alone.
- The three affected tests pass when run together.
- The entire `tests/extensions/background-terminals` directory passes with `--test-concurrency=1`.
- An unmodified `bun run test` reproduced the failure on the first run:
  - `kill settles a never-exiting process...` passed in about 3.32 s.
  - `concurrent overlapping multi-id kills...` failed in about 3.60 s.
  - `taskkill terminates a Windows descendant process tree` passed in about 3.62 s.
- The exact three-test selection then passed 3/3 at about 3.25-3.30 s each.
- Serial execution of the complete `background-terminals` directory passed: 89 passed, 4 skipped, 0 failed.

The failures only appear under full-suite cross-file concurrency. This rules out `.gitattributes`, line-ending policy, and the assertions in those three tests as the primary cause.

## Why this harms the whole repository

This is a test-infrastructure reliability problem, not a local inconvenience:

- **False CI failures block unrelated changes.** Any contributor can receive a red build even when their change does not touch background terminals or process management.
- **Flakes hide real regressions.** Once rerunning CI becomes normal, developers cannot tell whether a process-lifecycle failure is infrastructure noise or a genuine runtime bug.
- **Debugging time is wasted in the wrong layer.** The failure points at settlement, kill, or descendant assertions, encouraging investigation of production code and test assertions even though the trigger is suite-level scheduling.
- **Windows contributors pay a disproportionate cost.** The affected path uses real Windows process-tree semantics and `taskkill /T`, so the default full-suite topology makes their development loop less trustworthy.
[...]

PR #331, opened 2026-09-01
fix(workflows): enforce text projection line bounds

### LINKED ISSUE #329: Bug (Workflow): projectText ignores maxLines and duplicates input during line-triggered projection
## Summary

`projectText(content, { maxBytes, maxLines, recovery })` does not reliably enforce `maxLines`.

When projection is triggered only because the input exceeds the line limit—but remains well below the byte limit—the current implementation may select the entire input as both the head and the tail. The resulting projection duplicates the source text and can greatly exceed `maxLines` while still passing the final byte-only check.

Reproduced against upstream `main` at commit:

```text
5ad7c6e3b4f268a29413764a9fe10c7bb502d56b
```

## Minimal reproduction

```ts
import { Buffer } from "node:buffer";
import { projectText } from "./extensions/shared/text-projection.ts";

const content = Array.from(
  { length: 40 },
  (_, index) => `L${index}`,
).join("\n");

const projected = projectText(content, {
  maxBytes: 2048,
  maxLines: 3,
  recovery: "kept in artifacts",
});

console.log({
  inputBytes: Buffer.byteLength(content, "utf8"),
  inputLines: content.split("\n").length,
  outputBytes: Buffer.byteLength(projected, "utf8"),
  outputLines: projected.split("\n").length,
});

console.log(projected);
```

## Actual behavior

```text
inputBytes:  149
inputLines:  40
outputBytes: 416
outputLines: 85
```

The projected output is effectively:

```text
L0
...
L39

[... middle omitted ...]

L0
...
L39

[Projection bounded: showing 298 B of 149 B across the head and tail. kept in artifacts]
```

The complete 40-line input appears once as the head and again as the tail.

The footer also reports `showing 298 B of 149 B`, which demonstrates that the selected head and tail overlap.

The result satisfies the byte limit but violates the three-line limit by 82 lines.

## Expected behavior

`maxBytes` and `maxLines` should be simultaneous hard bounds:

```text
outputBytes <= maxBytes
AND
outputLines <= maxLines
```

For the reproduction above, a reasonable compact projection would be:

```text
L0
[... middle omitted ...]
L39
```

The exact formatting is less important than preserving these properties:

- the output does not exceed either configured limit;
- omitted content is indicated explicitly;
- head and tail selections do not overlap;
- useful beginning and ending context is retained when the budget permits;
- the exact result remains recoverable from Workflow artifacts.

## Root cause

The projection pipeline applies different constraints at different stages:

1. The initial `truncateHead` probe checks both `maxBytes` and `maxLines`.
2. Once truncation is required, head and tail selection uses byte-only helpers.
3. The final acceptance check verifies only `Buffer.byteLength(projected)`.
4. Therefore, a line-only overflow is detected initially but the line constraint is lost during transformation and final validation.

With a 2048-byte budget, the current 75%/25% split gives approximately:

```text
head budget: 1536 bytes
tail budget:  512 bytes
```

Because the complete input is only 149 bytes, it fits independently inside both budgets. Both selectors therefore return the complete input.

## Impact

`projectText` is currently used by several Workflow projections, including:

- completed Workflow results delivered to the parent model;
- background Workflow completion batches;
- upstream results appended to downstream Workflow prompts;
- expanded operator-facing completion reports.

This does not appear to corrupt the canonical Workflow result or its persisted audit artifacts. It affects bounded projections derived from that result.

Possible consequences include:

- model context unexpectedly exceeding the intended line budget;
- duplicated evidence influencing model interpretation;
- unnecessary context/token consumption;
- misleading projection telemetry;
- operator-facing reports exceeding their documented bounds.

The smallest reproduction uses `maxLines: 3`, but production call sites use larger limits such as 200, 400, and 600. Outputs containing many short lines—test summaries, file lists, logs, or compact records—can still trigger the same class of failure.

## Why this is easy to miss

Most large-output tests naturally exceed the byte limit as well as the line limit. In that situation, byte-based head/tail selection usually produces two distinct slices and appears correct.

The defect is exposed specifically by a different input shape:

```text
many lines + few bytes per line
```

This is a useful general testing lesson for APIs with multiple limits: each limit needs an independent test where only that dimension exceeds its bound.

A `{ maxBytes, maxLines }` contract is conjunctive. Both constraints must remain active during:

1. overflow detection;
2. budget allocation;
3. content selection;
4. framing/footer construction;
5. final validation.

Checking both limits only at the beginning is insufficient.

The invariant below would also have exposed the overlap in this case:

```text
reported source bytes shown <= total source bytes
```

## Acceptance criteria

- [ ] Content that fits within both limits is returned byte-for-byte unchanged.
- [ ] A projection caused only by `maxLines` does not exceed `maxLines`.
- [ ] A projection caused only by `maxBytes` does not exceed `maxBytes`.
- [ ] When both limits are exceeded, the result satisfies both limits.
- [ ] Head and tail selection does not duplicate overlapping source ranges.
- [ ] Framing lines, omission markers, footers, and multi-line recovery text are included in the line budget.
- [ ] Very small positive line budgets still produce an explicit bounded/omission representation.
- [ ] Zero-byte and zero-line budgets produce an empty projection.
- [ ] UTF-8 characters are not split or replaced with `�`.
- [ ] Tests cover LF, CRLF, ASCII, CJK text, and emoji.
- [ ] A regression test covers the exact 40-line, 149-byte reproduction above.

## Scope

This issue is about enforcing the existing projection contract.

It does not require:

- changing Workflow execution semantics;
- changing canonical result persistence;
- removing full audit artifacts;
- introducing a new global resource or artifact abstraction;
- changing the configured production limits themselves.

## Related issues

- Related to #90, which discusses bounded and fair Workflow result projection.
- Related to #117, which covers another boundary-condition failure in the same projection helper.
- This is not a duplicate of either issue: neither reports line-triggered projection ignoring `maxLines` and duplicating the input.

JuiXiang:
## Problem

Closes #329.

`projectText()` detects overflow with both `maxBytes` and `maxLines`, but the existing head/tail projection only applies byte limits. Its final validation also checks only the byte count.

A 40-line, 149-byte input projected with `maxLines: 3` and `maxBytes: 2048` therefore produces 85 lines and duplicates the complete input around the omission marker. The footer reports `showing 298 B of 149 B`, confirming that the selected head and tail overlap.

This affects bounded Workflow result messages, completion batches, handoffs, and operator-facing completion reports.

## Value

This change makes the `{ maxBytes, maxLines }` contract dependable. A caller that allocates 400 lines can treat that value as a hard upper bound rather than a hint that is checked once and then lost during projection. The byte limit cannot replace the line limit: logs, file lists, and compact records can contain hundreds of short lines while remaining well below their byte budget.

For model-facing Workflow results, enforcing both limits keeps projected output within the context shape selected by the caller. The current bug can turn a small line-only overflow into a larger result by including the same source text as both the head and the tail. Removing that overlap avoids spending context on duplicate evidence and leaves the available projection budget for distinct information from the beginning and end of the result.

This also improves the quality of the evidence presented to the model. Repeating the same conclusion or error text can give it disproportionate weight during synthesis, even though the repetition was introduced by the projection layer rather than the Workflow itself. A non-overlapping head/tail projection preserves more of the source's actual coverage and does not manufacture repetition.

The same helper is used for parent-model result delivery, background completion batches, downstream handoffs, and expanded operator reports. Fixing the shared projection contract gives those paths consistent behavior without adding separate truncation rules to each Workflow surface.

Finally, projection diagnostics become internally consistent. A footer such as `showing 298 B of 149 B` is impossible when the selected source ranges do not overlap, and the reported bounds once again describe the text that was actually returned.

The change is limited to derived projections. Workflow execution, canonical results, persisted audit artifacts, and configured production limits are unchanged. Content that already fits both limits still passes through byte-for-byte, while the exact source remains recoverable from the existing artifacts when projection is required.

## Approach

- Apply both byte and line budgets when selecting the projected head and tail.
- Preserve the existing 75/25 head-to-tail allocation.
- Include omission markers, separator lines, the footer, and multi-line recovery text in the line budget.
- Use a compact, visibly omitted representation when the line budget cannot fit the normal framed projection.
- Validate the completed projection against both limits before returning it.
- Preserve byte-for-byte passthrough for input that already fits both limits.
- Add regression coverage for line-only overflow, small budgets, UTF-8 input, LF/CRLF input, and combined byte/line limits.

## Validation

Passed:

- `git diff --check`
- Target-file Biome format check
- Target-file Biome lint with `--error-on-warnings`
- `bun run lint` — 280 files checked
- `tsc --noEmit`
- `node scripts/check-config-contract.mjs` — 14 persisted fields checked
- `node scripts/check-discipline-ledger.mjs` — 12 append-only rows checked
- Relevant Node tests:

  ```text
  node --test --experimental-strip-types \
    tests/extensions/shared/text-projection.test.ts \
    tests/extensions/workflows/prompt.test.ts \
    tests/extensions/workflows/handoff.test.ts
  ```

  Result: 40 passed, 0 failed.

Not completed cleanly in the existing Windows checkout:

- `bun run check` reaches `format:check`, which reports 262 formatting diagnostics caused by pre-existing CRLF working-tree files outside this PR. Both changed files pass the same formatter, and full lint plus type checking pass separately.
- `bun run test` reaches the existing Windows `background-terminals` process-termination failures and then stops making progress. Four failures were observed in that area before the run was interrupted. This matches #304; this PR does not modify background terminal code.

No benchmark was run because the change affects bounded text selection rather than a performance-sensitive execution path.

## Impact

- User-visible behavior: truncated Workflow reports stay within their configured line limit and no longer repeat overlapping source text.
- Model-visible context: bounded Workflow results and handoffs may contain a more compact projection when line limits are reached.
- Runtime and lifecycle: no change.
- Persisted configuration and data: no change.
- Compatibility: input that fits both limits remains unchanged. Exact truncated content remains recoverable from the existing Workflow audit artifacts.

### DIFF
--- a/extensions/shared/text-projection.ts
+++ b/extensions/shared/text-projection.ts
@@ -1,6 +1,13 @@
-import { formatSize, truncateHead } from "@earendil-works/pi-coding-agent";
+import {
+  formatSize,
+  truncateHead,
+  truncateTail,
+} from "@earendil-works/pi-coding-agent";
 
 const HEAD_SHARE = 0.75;
+const OMISSION_MARKER = "[... middle omitted ...]";
+// Three framing blanks, the omission marker, and the footer.
+const STATIC_PROJECTION_OVERHEAD_LINES = 5;
 
 function utf8Prefix(content: string, maxBytes: number) {
   const bytes = Buffer.from(content, "utf8");
@@ -14,16 +21,46 @@ function utf8Prefix(content: string, maxBytes: number) {
   return "";
 }
 
-function utf8Suffix(content: string, maxBytes: number) {
-  const bytes = Buffer.from(content, "utf8");
-  if (bytes.length <= maxBytes) return content;
-  let start = Math.max(0, bytes.length - maxBytes);
-  while (start < bytes.length) {
-    const value = bytes.subarray(start).toString("utf8");
-    if (!value.startsWith("�")) return value;
-    start += 1;
-  }
-  return "";
+function boundedHead(content: string, maxBytes: number, maxLines: number) {
+  const result = truncateHead(content, { maxBytes, maxLines });
+  if (result.content || !result.firstLineExceedsLimit) return result.content;
+  return utf8Prefix(content, maxBytes);
+}
+
+function compactProjection(
+  content: string,
+  maxBytes: number,
+  maxLines: number,
+) {
+  const marker =
+    maxBytes >= OMISSION_MARKER.length
+      ? OMISSION_MARKER
+      : ".".repeat(Math.min(3, maxBytes));
+  if (!marker || maxLines <= 0) return "";
+  if (maxLines === 1) return marker;
+
+  const bodyLineBudget = maxLines - 1;
+  const keepTail = bodyLineBudget >= 2;
+  const separatorBytes = keepTail ? 2 : 1;
+  const bodyByteBudget = maxBytes - marker.length - separatorBytes;
+  if (bodyByteBudget <= 0 || (keepTail && bodyByteBudget < 2)) return marker;
+
+  const headLines = keepTail
+    ? Math.max(1, Math.floor(bodyLineBudget * HEAD_SHARE))
+    : bodyLineBudget;
+  const tailLines = keepTail ? Math.max(1, bodyLineBudget - headLines) : 0;
+  const headBytes = keepTail
+    ? Math.max(1, Math.floor(bodyByteBudget * HEAD_SHARE))
+    : bodyByteBudget;
+  const tailBytes = keepTail ? Math.max(1, bodyByteBudget - headBytes) : 0;
+  const head = boundedHead(content, headBytes, headLines);
+  const tail = keepTail
+    ? truncateTail(content, {
+        maxBytes: tailBytes,
+        maxLines: tailLines,
+      }).content
+    : "";
+  return [head, marker, tail].filter(Boolean).join("\n");
 }
 
 /** Keep decision context at the start and verdict/evidence at the end. */
@@ -31,7 +68,7 @@ export function projectText(
   content: string,
   options: { maxBytes: number; maxLines: number; recovery: string },
 ) {
-  if (options.maxBytes <= 0) return "";
+  if (options.maxBytes <= 0 || options.maxLines <= 0) return "";
 
   const probe = truncateHead(content, {
     maxBytes: options.maxBytes,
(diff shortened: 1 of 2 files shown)


tt-a1i (approved):
Reviewed at exact head ded53f0. The change enforces the existing maxBytes/maxLines projection contract without altering canonical Workflow results or persisted artifacts. CI and focused boundary coverage are green; no blocking findings.

Issue #343, opened 2026-09-02
feat(web): expose Pi Trust state and native approval flows


sowen1023:
## Parent

- #76

## Problem

The Web runtime reuses Pi ProjectTrustStore when opening a workspace, but the browser does not expose the resulting trust/permission state or a native confirmation flow. Users cannot tell why an operation is unavailable, inspect the authority of the active workspace, or safely approve/deny an action without leaving the Web workbench.

This must not be solved with a browser-owned permission database or by silently widening the Session authority.

## Value

- Make the active security boundary visible before tools execute.
- Allow Web users to handle supported confirmations without bypassing Pi.
- Preserve one source of truth for Trust, approval, and tool authority.

## Approach constraints

- Pi Trust and confirmation mechanisms remain authoritative.
- A trust-changing request must bind to the exact canonical workspace and active Session.
- Only the authenticated controlling browser client may answer a pending confirmation.
- Confirmation receipts must distinguish approved, denied, expired, stale, and already-settled requests.
- Snapshot and SSE events must be bounded and must not expose credentials or private confirmation payloads.
- Do not persist a second browser-specific permission policy.
- Child and package tool boundaries must remain fail-closed.

## Acceptance criteria

- [ ] The Web snapshot reports the active workspace Trust state from Pi-owned state.
- [ ] The UI clearly displays trusted, untrusted, restricted, and unknown states that Pi can distinguish.
- [ ] Unsupported operations explain the relevant Trust/permission boundary instead of failing silently.
- [ ] Supported pending confirmations can be inspected and approved or denied through the native confirmation seam.
- [ ] Stale, cross-workspace, cross-Session, and non-controller responses are rejected.
- [ ] Refresh/reconnect restores the canonical pending/settled state without replaying an approval.
- [ ] No Web action can grant authority beyond the Pi Session or package policy.
- [ ] Tests cover untrusted workspaces, approval, denial, expiry, stale receipts, reconnect, and shutdown.
- [ ] Documentation and setup status describe the Web Trust surface.
- [ ] bun run check and bun run test pass.

## Out of scope

- Remote/public authentication.
- A second ACL system.
- Automatic project trust.

testikun:
[...]

PR #360, opened 2026-09-03
feat(web): project Pi trust state in snapshots

seekskyworld:
## Problem

Implements the snapshot portion of #343. Web clients could not distinguish the active workspace trust state before invoking operations.

## Value

The browser can show whether the current workspace is trusted without owning or persisting a second permission policy.

## Approach

Read Pi SettingsManager trust state through the Web runtime, map it to bounded `trusted`/`untrusted`/`unknown` values, and include it in the existing runtime snapshot projection. Unbound or unreadable state fails closed to `unknown`.

## Validation

- `npx tsc --noEmit`
- `node --test --experimental-strip-types tests/web/protocol.test.ts tests/web/pi-adapter.test.ts tests/web/pi-runtime.test.ts` (44 passed)
- `git diff --check`

## Impact

- User-visible behavior: snapshot consumers receive trust status for display.
- Runtime/lifecycle: reads Pi-owned trust state only.
- Compatibility/risk: optional controller field preserves existing test doubles; approval actions remain a separate follow-up.

### DIFF
--- a/web/adapter/pi-adapter.ts
+++ b/web/adapter/pi-adapter.ts
@@ -490,6 +490,7 @@ export class PiWebAdapter {
         status: this.runtime.isIdle()
           ? ("idle" as const)
           : ("running" as const),
+        trust: this.runtime.projectTrustState ?? "unknown",
         capabilities: webCapabilitySnapshot(this.runtime.sessionManager),
       },
       truncation: {
--- a/web/protocol/types.ts
+++ b/web/protocol/types.ts
@@ -113,6 +113,7 @@ export interface WebSnapshot {
   models: WebModelSummary[];
   runtime: {
     status: "idle" | "running" | "unknown";
+    trust: "trusted" | "untrusted" | "unknown";
     capabilities: WebCapabilitySnapshot;
   };
   truncation: WebSnapshotTruncation;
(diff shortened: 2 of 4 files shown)


tt-a1i (commented):
## 请让给 #392

`runtime.trust` 来自 Pi `isProjectTrusted()`，缺了 fail-closed 成 `unknown`，方向对。但把 restricted/undecided 压成 `untrusted`，字段也不如 #392 完整，而且没有测试。

请关掉，改合 #392。Node 24 的 `tests/web/cli.test.ts` 红也请先分清是不是共享抖动。

tt-a1i (commented):
Exact-head review found no new P2+ in the narrow original projection; focused protocol/adapter/runtime tests passed 44/44. Relative to current main, however, the value is already covered by #442’s richer Pi-owned getProjectTrustStatus/trustInfo and GET /api/trust projection, including stored decision, restricted/unknown state and refresh-required facts. Recommend treating this proposal as superseded rather than adding a parallel reduced trust field. The branch also conflicts with main and has a historical Node 24 CLI smoke failure; the latter is not established as a regression from this diff. No merge or closure performed.

tt-a1i:
Closing as superseded by the richer Pi-owned trust projection integrated in #442. Current main preserves stored decision, restricted/unknown state, resource provenance and refresh-required facts; adding this older parallel trust field would duplicate that source of truth. Thank you for the proposal and implementation. The useful unarchive work in #361 is being handled separately with author attribution.

Issue #349, opened 2026-09-02
feat(web): add bounded transcript-aware session search


sowen1023:
## Parent

- #76

## Problem

Web search currently filters only the bounded client snapshot by Session title and cwd. It does not search transcript content, tool evidence, archived Sessions, or Sessions omitted from the current snapshot limit.

As the number and length of Sessions grow, users cannot reliably recover prior decisions, errors, commands, or code changes.

## Value

- Find relevant prior work without loading every Session into the browser.
- Keep search useful beyond the current snapshot window.
- Preserve Session files as source of truth while treating the search index as derived and recoverable.

## Approach constraints

- Search must be server-side, bounded, cancellable, and workspace-aware.
- A cache/index may accelerate lookup but cannot become authoritative Session storage.
- Results retain Session, message/turn, workspace, source, timestamp, and snippet provenance.
- Enforce query length, result count, scanned bytes/files, concurrency, and time limits.
- Respect archive visibility, workspace roots, trust, and Web-vs-terminal Session access rules.
- Snippets must be safely escaped and explicitly marked when truncated.
- Corrupt, changing, or deleted Session files must degrade without poisoning the index.
- Do not send private transcript content to an external search service.

## Acceptance criteria

- [ ] Search covers titles, cwd, user/assistant text, and bounded tool evidence where allowed.
- [ ] Results can include archived and compatible read-only Sessions through explicit filters.
- [ ] Result pagination/cursors remain stable enough for navigation and report invalidation.
- [ ] Every result links to the exact Session and bounded turn/message context.
- [ ] Query, time, byte, file, concurrency, and result limits are enforced and observable.
- [ ] Index rebuild/invalidation follows Session create, rename, append, archive, unarchive, delete, and external file changes.
- [ ] No result crosses an unauthorized workspace or Session source boundary.
- [ ] Large, malformed, Unicode, and adversarial transcripts are covered by tests.
- [ ] The current lightweight title/cwd filtering remains available while server search is pending.
- [ ] bun run check and bun run test pass.

## Out of scope

- External/cloud semantic search.
- Making the search index a canonical memory system.
- Unbounded full-history replay to the browser.

testikun:
我来认领 #349 中尚未覆盖的“有界 transcript 内容扫描器”子范围，不重复 PR #362 的 metadata endpoint 和 PR #376 的 offset 分页。

技术方案：
[...]

PR #362, opened 2026-09-03
feat(web): add bounded session search endpoint

seekskyworld:
## Problem

Implements the bounded metadata-search slice of #349. Web search was limited to browser-side filtering and had no reusable server boundary.

## Value

Clients can search Session names, first-message previews, and workspace paths without loading every transcript into the browser.

## Approach

Add an adapter search method with a 200-character query bound, optional archived filtering, and a 100-result cap. Expose it through authenticated `GET /api/search`; canonical Session files remain the source of truth and transcript-aware indexing remains future work.

## Validation

- `npx tsc --noEmit`
- `node --test --experimental-strip-types tests/web/web-host.test.ts tests/web/pi-adapter.test.ts` (32 passed)
- `git diff --check`

## Impact

- User-visible behavior: adds a bounded server search endpoint.
- Runtime/lifecycle: read-only projection.
- Compatibility/risk: no existing endpoint changes; full transcript search/indexing is not claimed.


tt-a1i (commented):
## 和 #376 选一个；这不是 #387

内存里搜 name/preview/cwd，200 字符 / 100 命中，不加载 transcript，归档默认不搜 — 切片干净。但没有分页、没有测试、`truncated` 写死。#376 改了同一套函数的参数形状。

请和 #376 合成一份带测试的元数据搜索。transcript 正文搜索留给 #387。

tt-a1i (changes requested):
Exact-head review: 96a71f6d1be15bfc0a43053045144acdaaaec5a7

[P1] Server-side search cannot find sessions outside the newest 500-entry UI projection. searchSessions() calls listSessionProjection(), which first truncates SessionManager.listAll() to WEB_MAX_SESSIONS at web/adapter/pi-adapter.ts:314-340, and only then applies the query at lines 514-527. With 501 sessions, a unique match in the oldest session is guaranteed to return no result even though the canonical Session metadata exists.

That defeats the central value of moving search to a reusable server boundary and makes missing results look authoritative. Please search the canonical Pi-owned metadata before applying the response cap, while keeping query/result/work bounds explicit; add a regression with more than 500 sessions whose only match is beyond the snapshot projection.

Verification boundary: exact-head static path analysis and boundary construction; current tests cover result bounds but not a match outside the projection cap.

seekskyworld:
Rebased onto current `main` and addressed the P1:

- `searchSessions()` now walks canonical `SessionManager.listAll()` metadata rather than the 500-entry UI projection.
- Query bound remains 200 characters; default page is 100; transcripts are not loaded; archived sessions are excluded unless `archived=true`.
- Added a regression with more than 500 sessions whose only match is outside the snapshot projection.
- Empty queries are 400 rather than a silent empty 200.

Pagination (`offset`/`limit`/`nextOffset`) is on #376, rebased onto this head. This is metadata search, not #387 transcript search.

tt-a1i (commented):
审查提交：`c54b92012930307629345e82fd6bf9f1c74f824d`。

建议先整合重叠方案，再提交最终实现。

### 验证范围

Complete diff reviewed. Old 500-item projection bug fixed; 2 exact-head adapter/HTTP tests pass. Host still deliberately exposes capped first page without nextOffset; #376 supplies pagination.

### 限制与后续

Prefer #376 consolidation. PR prose says /api/search while implementation is /api/sessions/search. Canonical scan remains full and uncancelled; no performance claim.

PR #364, opened 2026-09-04
fix(web): preserve running state for queued prompts

seekskyworld:
## Problem

Closes the state-regression slice of #363. When a follow-up prompt was accepted during an active turn, the Web UI unconditionally changed the status from running back to preparing.

## Value

Long workflow/subagent turns retain an accurate running indicator while queued prompts wait for admission.

## Approach

Centralize prompt-accepted state handling and preserve `running` when already established; only new work enters `preparing`, and settled prompts return to idle.

## Validation

- `node --test --experimental-strip-types tests/web/app-render.test.ts`
- `git diff --check`

## Impact

- User-visible behavior: queued prompts no longer regress the status label.
- Runtime/lifecycle: presentation-only state correction.
- Compatibility/risk: no protocol changes.


tt-a1i (commented):
## 切片正确，但会被 #370 盖掉

已经 `running` 时保持 running、不从 UI 推断完成，这是对的。没有测试。#370 会改同一批行。请并进 #370（或 #384 的 store），不要单独合，否则后合的会把这份修掉。

Issue #346, opened 2026-09-02
feat(web): add inspectable and controllable OpenPI capability panels


sowen1023:
## Parent

- #76

## Problem

The first Web workbench projects compact status summaries for Subagents, Workflows, and Background Terminals. The browser cannot inspect their full bounded evidence, understand dependency/state transitions, or invoke supported lifecycle controls. Tasks/Goals have no Web projection yet.

This limits the workbench to passive status badges for the OpenPI capabilities that most need long-running operational visibility.

## Value

- Give operators one place to inspect delegated and background work.
- Expose exact progress and terminal evidence without copying capability logic into the browser.
- Allow supported stop/cancel actions through the capability owner.

## Approach constraints

- Existing capability registries and lifecycle events remain authoritative.
- The browser consumes stable projections; it must not reconstruct completion from labels.
- Controls are exposed only where the owning capability already has exact semantics and authority checks.
- Every action binds resource identity, Session identity, controller lease, and current epoch.
- Logs, graphs, node lists, results, and retained resources require explicit size and retention bounds.
- Tasks/Goals must follow their accepted Pi/OpenPI lifecycle design rather than introducing a Web-only task system.

## Acceptance criteria

- [ ] Subagent panels show identity, role/model when available, status, elapsed time, bounded progress, result, and cancellation evidence.
- [ ] Workflow panels show graph/node state, dependencies, bounded input/output evidence, failure, replay/provenance, and terminal result.
- [ ] Background Terminal panels show command identity, bounded output, readiness, exit/cancel state, and supported stop controls.
- [ ] Supported lifecycle commands return exact accepted/stale/settled/failed receipts and are idempotent.
- [ ] Tasks/Goals appear only after their canonical runtime contract exists, using the same projection boundary.
- [ ] Snapshot, SSE replay, reconnect, Session switch, and shutdown converge without ghost resources.
- [ ] Capability details remain bounded and provide recovery references for omitted evidence.
- [ ] Read-only viewers cannot invoke controller actions.
- [ ] Focused registry/protocol/UI/browser tests, bun run check, and bun run test pass.

## Related work

- #43 tracks the broader Tasks design.
- #161 tracks global safe pause semantics.
- #196 tracks cross-capability lifecycle ownership and cancellation documentation.

## Out of scope

- A second orchestration runtime.
- Browser-owned task/workflow state.
- Recursive team management.

testikun:
[...]

PR #366, opened 2026-09-04
feat(web): expose bounded capability snapshot endpoint

seekskyworld:
## Problem

Implements the read-only protocol slice of #346. The Web workbench had capability data only inside the full snapshot, with no focused endpoint for capability panels and reconnect refreshes.

## Value

Operators can request the existing bounded Subagent/Workflow/Terminal projections directly without the browser reconstructing lifecycle state.

## Approach

Expose authenticated `GET /api/capabilities`, using the existing `webCapabilitySnapshot` projection and active Session identity. No new registry or control path is introduced.

## Validation

- `npx tsc --noEmit`
- `git diff --check`

## Impact

- User-visible behavior: enables focused capability-panel refreshes.
- Runtime/lifecycle: read-only projection only.
- Compatibility/risk: additive endpoint; lifecycle controls remain owner-specific follow-ups.

### DIFF
--- a/web/host/web-host.ts
+++ b/web/host/web-host.ts
@@ -9,7 +9,10 @@ import {
 } from "node:http";
 import { URL } from "node:url";
 import { promisify } from "node:util";
-import { subscribeWebCapabilities } from "../../extensions/shared/web-observer-registry.ts";
+import {
+  subscribeWebCapabilities,
+  webCapabilitySnapshot,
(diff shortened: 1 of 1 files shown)


tt-a1i:
This exact capabilities endpoint was integrated through #367 (main commit da751945). The merged commit retains seekskyworld as its original author and includes the original endpoint commit in its message. Closing the already integrated PR while preserving this source record.

Issue #348, opened 2026-09-02
feat(web): add Pi-native provider setup, reasoning controls, and OpenPI settings


sowen1023:
## Parent

- #76

## Problem

The Web model picker can select an already available provider/model pair, but the browser cannot:

- inspect provider authentication availability without leaking credentials;
- start or complete a supported Pi-native provider login flow;
- view or set the current thinking/reasoning level;
- enter the canonical OpenPI setup episode;
- inspect package setup status or runtime provenance.

Users must leave the Web workbench to understand or change these settings, and unavailable models provide limited recovery guidance.

## Value

- Make model configuration usable from the browser while keeping Pi authoritative.
- Preserve the single OpenPI setup entry point and typed persistence boundary.
- Give users enough provenance to diagnose which runtime/configuration they are operating.

## Approach constraints

- Pi remains the source of truth for provider credentials, available models, and reasoning state.
- Credentials, tokens, and secrets must never enter snapshots, SSE events, browser storage, or logs.
- Model and reasoning changes bind to the expected Session and return exact accepted/stale/failed receipts.
- OpenPI configuration must continue through the canonical /openpi-setup episode and episode-scoped typed writer.
- Do not add a browser parser, direct preference-file writes, or an extension-specific setup command.
- Package-owned settings must keep setup config, status output, SETUP.md, README defaults, and tests synchronized.
- Authentication UI must use a reviewed Pi/provider seam; unsupported providers get explicit external instructions.

## Acceptance criteria

- [ ] Web reports bounded provider/model availability and authentication status without secret material.
- [ ] Supported provider login uses Pi-owned authentication and reports success, denial, expiry, and failure.
- [ ] Users can inspect and change the thinking/reasoning level supported by the selected model.
- [ ] Model plus reasoning updates are atomic for the expected active Session.
- [ ] Web exposes the canonical /openpi-setup flow rather than a second configuration path.
- [ ] No-argument setup status and Web status agree on package-owned values.
- [ ] A read-only diagnostics view shows OpenPI version, checkout/package source, Pi version, active Session source, and relevant bounded capability provenance.
- [ ] Reconnect cannot repeat a login or configuration write.
- [ ] Tests prove secrets never enter serialized protocol state.
- [ ] Setup/config documentation and drift guards are updated.
- [ ] bun run check and bun run test pass.

## Related work

- #195 defines provider/model/fallback/reasoning state and invocation evidence.

## Out of scope

- A second provider stack.
[...]

PR #367, opened 2026-09-04
feat(web): expose bounded runtime diagnostics

seekskyworld:
## Problem

Implements the read-only diagnostics slice of #348. Web users had no bounded way to inspect the active runtime, Session identity, workspace selection, or current model.

## Value

Support can distinguish runtime and Session configuration without exposing credentials or transcript data.

## Approach

Add authenticated `GET /api/diagnostics` returning Node version, canonical runtime cwd, active Session id, workspace selection, and current model summaries. Provider credentials and secrets remain outside the projection.

## Validation

- `npx tsc --noEmit`
- `git diff --check`

## Impact

- User-visible behavior: adds a bounded diagnostics response for the Web workbench.
- Runtime/lifecycle: read-only inspection.
- Compatibility/risk: additive endpoint; provider login and setup episode remain separate work.


tt-a1i (approved):
Approved at exact head `3bd6dfc9ecf2604c42c59c8f78356d8a48e7eeb1`. I independently reviewed the scoped implementation and boundary conditions, verified the current diff against main, and found no unresolved P0/P1 issues. Required CI checks are green at this head.

PR #368, opened 2026-09-04
refactor(git-read): use shared search seams

seekskyworld:
## Problem

Implements the cross-extension import slice of #73. `git-read` directly imported private `file-search/src` modules, coupling extension internals.

## Value

A named shared seam makes the dependency direction explicit and gives future refactors one stable import boundary without changing command, output, or process semantics.

## Approach

Add shared re-export seams for bounded search output and process execution, then update `git-read` to consume only those shared modules. The underlying implementation and behavior remain unchanged.

## Validation

- `npx tsc --noEmit`
- `git diff --check`

## Impact

- Runtime/lifecycle: no behavior change.
- Compatibility/risk: additive shared modules; remaining #73 items are intentionally out of scope.

### DIFF
--- a/extensions/file-search/index.ts
+++ b/extensions/file-search/index.ts
@@ -42,7 +42,10 @@ import {
   type PlatformTarget,
   type ResolvedBinary,
 } from "./src/binaries.ts";
-import { formatCapturedOutput, type CapturedOutput } from "./src/output.ts";
+import {
+  formatCapturedOutput,
+  type CapturedOutput,
+} from "../shared/search-output.ts";
 import {
   FD_PARAMETER_DESCRIPTIONS,
   FD_PROMPT_GUIDELINES,
@@ -53,7 +56,10 @@ import {
   RG_PROMPT_SNIPPET,
   RG_TOOL_DESCRIPTION,
 } from "./src/prompt.ts";
-import { discardCapturedOutput, executeSearchProcess } from "./src/process.ts";
+import {
+  discardCapturedOutput,
+  executeSearchProcess,
+} from "../shared/search-process.ts";
 import {
   OPENPI_TOOL_SURFACE,
   patchOwnedTools,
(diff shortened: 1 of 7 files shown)


tt-a1i (changes requested):
Exact-head review: a9d7ec493ccf2277768a7b16dafa6bb02719a2a1

[P1] This moves the import name but does not remove the cross-extension private dependency required by Issue #73. extensions/shared/search-output.ts and search-process.ts are only re-export shims whose implementations still come from ../file-search/src/output.ts and ../file-search/src/process.ts. The shared layer therefore depends on a feature extension private src tree; refactoring or packaging file-search can still break git-read through the new alias.

Issue #73 acceptance requires extensions to stop importing another extension private implementation, and the PR claims a stable shared seam. Please move the reusable bounded-output/process implementation into extensions/shared (or define a genuinely owner-neutral shared module), then have both file-search and git-read depend on it. Add an import-boundary guard so a future shared-to-feature dependency fails CI.

Verification boundary: exact-head Standards + Spec static review; no runtime behavior change is claimed, so the blocker is the unresolved dependency direction itself.

seekskyworld:
Rebased onto current `main` and addressed the P1:

- Moved the bounded output/process implementation into `extensions/shared/search-output.ts` and `search-process.ts`.
- `file-search` and `git-read` now depend on that shared module; there is no re-export shim back into `file-search/src`.
- Added `tests/extensions/shared/import-boundary.test.ts` so a future shared→feature import fails CI.

tt-a1i (commented):
审查提交：`d0572d0758e9c313f1ed1c893e85f5fa473f90db`。

### 验证范围

Shared implementation is now physically moved, not re-exported from feature internals: old architecture finding resolved. Full rename-aware diff reviewed. 14 Node tests and 30 file-search Vitest tests pass.

### 限制与后续

Needs current-main integration and refreshed final-head checks. No native Windows runtime run locally.

PR #373, opened 2026-09-04
feat(subagents): add structured direct results

testikun:
## Problem

Direct Subagents can currently return only free-form assistant text. Callers that need typed review findings, research evidence, or test matrices must parse prose, while Workflow already has a private structured-result mechanism that Direct Subagents cannot reuse.

Related to #155. Does not close the Issue. Remaining #155 work: a text-vs-typed success-rate and token comparison; an operator-facing writeup of every terminal state.

## Value

Callers can opt into a schema-validated Direct Subagent result without changing the default text path. The result is delivered as typed tool details and retained as a private, immutable content-addressed JSON artifact, giving downstream work an exact machine-readable handoff.

## Approach

- add optional `output_schema` to `subagent_spawn`;
- extract Workflow's terminating `structured_output` tool into a shared child-safe primitive, preserving Workflow's existing artifact-budget behavior;
- inject the tool and instruction only for opted-in runs, and fail closed if the child finishes without a valid submission;
- bound structured schemas and values, persist complete JSON with a SHA-256 identity and mode `0600`, and expose the structured value/artifact path through automatic delivery and explicit waits;
- keep ordinary text-only Direct Subagent sessions unchanged.

## Validation

- `bun run check` — passed.
- full Node 22 suite via `node scripts/run-tests.mjs` — 1,248 passed, 1 skipped, 0 failed.
- Vitest suite — 30 passed, 0 failed.
- focused Workflow compatibility and shared structured-output tests — 22 passed, 0 failed.

## Impact

- User-visible behavior: `subagent_spawn` accepts an optional `output_schema`; no UI changes.
- Model-visible context/tools: opted-in children receive one terminating `structured_output` tool and one short instruction; default children do not.
- Runtime/lifecycle: an opted-in run succeeds only after a valid structured submission and durable artifact write.
- Persisted config/data: no config change; successful structured results add private content-addressed `.json` artifacts alongside existing subagent result artifacts.
- Compatibility/risk: additive for Direct Subagents; the refactor preserves Workflow's prior schema tool, limits, and failure semantics.


tt-a1i (approved):
Approved at exact head `5b9d36d1fed3c3f62108b2ce3f428211aa5c6228`. I independently reviewed the scoped implementation and boundary conditions, verified the current diff against main, and found no unresolved P0/P1 issues. Required CI checks are green at this head.

PR #376, opened 2026-09-04
feat(web): bound session search pagination

seekskyworld:
## Problem

Adds the pagination slice of #349. A bounded server search needs a stable way to retrieve additional results without loading an unbounded list.

## Value

Clients can request 1-100 metadata results at a time with an explicit next offset.

## Approach

Add validated `offset` and `limit` query parameters to the read-only Session search endpoint and adapter projection. Existing query, archive, and 200-character bounds remain enforced.

## Validation

- `npx tsc --noEmit`
- `git diff --check`

## Impact

- User-visible behavior: paged search responses.
- Runtime/lifecycle: read-only projection.
- Compatibility/risk: additive endpoint; transcript indexing remains out of scope.


tt-a1i (commented):
## 和 #362 冲突；不要当成 #387

分页上界（1–100、host 拒不安全 offset/limit）是对的，但：

- 和 #362 的 `searchSessions` 参数个数 / `nextOffset` vs `truncated` 不兼容。两份里只留一份，或先 rebase。
- 空 `q` 返回 200 + `[]`。最后一页长度刚好等于 limit 时会多出一个假 `nextOffset`。
- 没有测试。
- 这是元数据搜索，不是 #387 的 transcript 正文搜索。

tt-a1i (changes requested):
Exact-head review: e2022019fe16649c461373ff4de0c024ba81fd64

[P1] Pagination is applied after the same 500-session UI projection, so it cannot retrieve later canonical results. web/adapter/pi-adapter.ts:314-340 retains at most WEB_MAX_SESSIONS before searchSessions() filters and slices at lines 513-522. Any offset at or beyond 500 is empty, and even earlier pages omit matches that fell outside the newest 500 sessions. The endpoint can still emit a nextOffset as if more canonical search results were reachable.

This breaks the PR goal of stable bounded retrieval without loading an unbounded list. Please paginate a canonical or purpose-built bounded search traversal rather than the snapshot projection, and return continuation evidence derived from that traversal. Add tests with more than 500 sessions, including a unique late match and an offset beyond 500.

Verification boundary: exact-head static boundary analysis; current validation does not exercise data beyond the projection cap.

seekskyworld:
Rebased onto #362 / current `main`:

- Pagination is applied to the canonical metadata match list, not the UI projection.
- Host accepts bounded `offset`/`limit` (limit 1–100) and emits `nextOffset` only when later matches remain.
- Added coverage for offset 500 with a remaining late match, and 400s for unsafe offset/limit.

Keep #387 for transcript content search.

tt-a1i (commented):
审查提交：`ae3b1fb1b2fd96ee278bc72e4fdd542d2ed3e57b`。

建议先整合重叠方案，再提交最终实现。

### 验证范围

Complete diff reviewed. Adapter searches canonical SessionManager.listAll before slicing; old 500-item cap bug resolved. 2 exact-head targeted adapter/HTTP tests pass including 501st session and offset/limit validation.

### 限制与后续

Response is bounded, canonical scan is not work-bounded or cancellable; no large-history stress evidence. Superset of #362; choose one implementation. No UI integration claimed.

Issue #344, opened 2026-09-02
feat(web): add bounded file and image attachments with safe rendering


sowen1023:
## Parent

- #76

## Problem

The Web composer currently accepts only a string prompt. It has no file picker, paste/drop attachment flow, image preview, or multimodal prompt projection. Model output also lacks a reviewed browser policy for rendering image content, so remote image references remain inert.

This prevents common coding workflows such as attaching a screenshot, a log, or a small reference file while preserving the original evidence identity.

## Value

- Support screenshot-driven debugging and multimodal models.
- Let users attach bounded evidence without copying large content into the text box.
- Make input provenance, model compatibility, and cleanup explicit.

## Approach constraints

- Reuse Pi message/content-block semantics and provider capability checks.
- Bind uploaded content to the expected workspace, Session, and prompt command.
- Enforce explicit count, per-file, aggregate-byte, MIME, and filename bounds before admission.
- Do not turn a browser-supplied path into unrestricted filesystem authority.
- Keep canonical bytes separate from bounded model/UI projections and report truncation honestly.
- Clean temporary artifacts on rejection, cancellation, settlement, and host shutdown.
- Render untrusted SVG/HTML and remote resources under a reviewed browser security policy.

## Acceptance criteria

- [ ] Users can select, paste, or drop supported files and images into the composer.
- [ ] The UI shows name, type, size, removal state, and admission errors before sending.
- [ ] Unsupported models/providers fail clearly without silently dropping attachments.
- [ ] Prompt admission is atomic: text and all accepted attachments target the same expected Session.
- [ ] Count, MIME, filename, per-file, and aggregate size limits are tested.
- [ ] Traversal, symlink, stale-session, cancellation, and cleanup cases fail closed.
- [ ] Supported image output is rendered safely; unsupported content has a usable download/link fallback.
- [ ] Refresh/reconnect does not duplicate an admitted attachment.
- [ ] Protocol, host, runtime, UI, and real-browser tests cover the flow.
- [ ] bun run check and bun run test pass.

## Out of scope

- Cloud attachment storage.
- Arbitrary large-file transfer.
- Fetching remote URLs on behalf of the user without a separate permission design.

testikun:
[...]

PR #365, opened 2026-09-04
feat(web): add bounded attachment admission rules

seekskyworld:
## Problem

Implements the admission-contract slice of #344. Web attachments had no shared, testable bounds for count, filename, MIME, per-file size, or aggregate size.

## Value

Future file/image upload paths can fail closed before touching workspace or Session authority.

## Approach

Add a pure protocol validator with explicit supported MIME types, filename traversal protection, count and UTF-8-independent byte limits. No upload storage or prompt behavior is changed.

## Validation

- `node --test --experimental-strip-types tests/web/attachments.test.ts`
- `npx tsc --noEmit`
- `git diff --check`

## Impact

- User-visible behavior: establishes bounded attachment admission errors for future UI.
- Model-visible context/tools: none.
- Runtime/lifecycle: none.
- Persisted config/data: none.
- Compatibility/risk: additive protocol utility; upload flow remains unimplemented.

PR #380, opened 2026-09-04
fix(web): validate attachment filename types

seekskyworld:
## Problem

Adds the MIME/filename consistency slice of #344. Attachment admission could accept a filename whose extension contradicted its declared content type.

## Value

Future upload paths get a clearer fail-closed boundary against misleading or malformed attachment metadata.

## Approach

Require supported MIME types to use an allowed filename extension while retaining count, traversal, per-file, and aggregate byte limits.

## Validation

- `npx tsc --noEmit`
- `git diff --check`

## Impact

- User-visible behavior: invalid type/extension pairs are rejected.
- Runtime/lifecycle: none.
- Compatibility/risk: additive validation for future attachment admission.

### DIFF
--- a/web/protocol/attachments.ts
+++ b/web/protocol/attachments.ts
@@ -1,0 +1,71 @@
+export const WEB_MAX_ATTACHMENTS = 8;
+export const WEB_MAX_ATTACHMENT_BYTES = 2 * 1024 * 1024;
+export const WEB_MAX_ATTACHMENT_TOTAL_BYTES = 8 * 1024 * 1024;
+
+const SUPPORTED_MIME = new Set([
+  "text/plain",
+  "text/markdown",
+  "application/json",
+  "image/png",
+  "image/jpeg",
+  "image/webp",
+]);
+
+export interface WebAttachmentInput {
+  readonly name: string;
+  readonly mime: string;
+  readonly size: number;
+}
+
+function isSafeAttachmentName(name: string) {
+  if (name.length < 1 || name.length > 120) return false;
+  if (name === "." || name === "..") return false;
+  if (name.includes("..") || name.includes("/") || name.includes("\\")) {
+    return false;
+  }
+  return !/[\u0000-\u001f\u007f]/u.test(name);
+}
+
+export function validateWebAttachments(
+  attachments: readonly WebAttachmentInput[],
+) {
+  if (attachments.length > WEB_MAX_ATTACHMENTS) {
+    return {
+      ok: false as const,
+      error: `at most ${WEB_MAX_ATTACHMENTS} attachments are allowed`,
+    };
+  }
+  let total = 0;
+  for (const attachment of attachments) {
+    if (!isSafeAttachmentName(attachment.name)) {
+      return {
+        ok: false as const,
+        error: `invalid attachment name: ${attachment.name}`,
+      };
+    }
+    if (!SUPPORTED_MIME.has(attachment.mime)) {
+      return {
+        ok: false as const,
+        error: `unsupported attachment type: ${attachment.mime}`,
+      };
+    }
+    if (
+      !Number.isSafeInteger(attachment.size) ||
+      attachment.size < 0 ||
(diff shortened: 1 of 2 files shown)


tt-a1i (commented):
## 切片对，但会和 #365 硬冲突

MIME↔扩展名白名单、拒绝 `..`，方向对。问题：

1. 整文件重造了 `attachments.ts`，和 #365 合不了。请 rebase 到 #365 上，只留类型校验。
2. 没有测试。`toLocaleLowerCase()` 做扩展名折叠，会吃 locale。
3. #388 的 staging 现在还没调用这个校验。准入应先于暂存。

web/protocol/attachments.ts
```diff
@@ -0,0 +1,22 @@
...
+  "image/jpeg": [".jpg", ".jpeg"], "image/webp": [".webp"],
+};
+export interface WebAttachmentInput { readonly name: string; readonly mime: string; readonly size: number; }
+export function validateWebAttachments(attachments: readonly WebAttachmentInput[]) {
+  if (attachments.length > WEB_MAX_ATTACHMENTS) return { ok: false as const, error: "too many attachments" };
+  let total = 0;
+  for (const attachment of attachments) {
+    if (!/^(?!\.\.?(?:$|\.))[\w .()\[\]-]{1,120}$/u.test(attachment.name) || attachment.name.includes("..")) return { ok: false as const, error: "invalid attachment name" };
+    if (!MIME_EXTENSIONS[attachment.mime]) return { ok: false as const, error: "unsupported attachment type" };
+    if (!MIME_EXTENSIONS[attachment.mime].some((extension) => attachment.name.toLocaleLowerCase().endsWith(extension))) return { ok: false as const, error: "attachment extension does not match type" };
```
tt-a1i, line 16: [P2] Reject inherited object keys before invoking the extension list

MIME_EXTENSIONS is an ordinary object, so a MIME value of `constructor` or `__proto__` passes the truthiness check on the preceding line. This line then throws `TypeError: MIME_EXTENSIONS[attachment.mime].some is not a function`, rather than returning `{ ok: false, error: "unsupported attachment type" }`. I reproduced both with the exact-head module and `validateWebAttachments([{ name: "x.txt", mime, size: 1 }])`; the normal `text/plain` case succeeds.

Use an own-property whitelist lookup (or a Map/null-prototype record) and add regression cases for inherited keys so arbitrary unsupported MIME strings are rejected without throwing. This is the pure admission contract, not evidence of an active upload endpoint.

tt-a1i (changes requested):
Exact-head review of 5bf88a1bd71e2dfd5875e74394e6d9c30b43891d. The MIME/extension admission boundary is useful, but malformed MIME metadata currently throws instead of returning the validator's bounded rejection result. The existing integration concerns with #365/#388 also remain; this review adds one independently reproduced correctness finding. No deployed HTTP or installed runtime impact is claimed.

seekskyworld:
Rebased onto #365 / current `main` and addressed the exact-head finding:

- MIME whitelist is now a `Map`, so `constructor` / `__proto__` / `toString` return `{ ok: false, error: "unsupported attachment type" }` instead of throwing.
- Extension matching uses ASCII `toLowerCase()`, not `toLocaleLowerCase()`.
- Added regressions for inherited MIME keys and MIME/extension mismatch.
- This PR now contains only the type-check delta on top of the #365 admission contract.

Staging (#388) still needs to call this validator before writing bytes; this slice does not duplicate that lifecycle.

tt-a1i (commented):
审查提交：`19f01fa136fa3226f064babe951713d07d737a71`。

建议先整合重叠方案，再提交最终实现。

### 验证范围

Complete validator/test diff reviewed. Map-based MIME lookup fixes prior inherited-key exception; 4 tests pass. Production grep finds only definition, no upload caller.

### 限制与后续

Pure admission helper, not an implemented upload flow. Superset of #365; integrate one shared validator with actual staging path rather than merge both independently.

PR #352, opened 2026-09-02
feat(web): restore polished workbench UI and conversation rendering

somewan820:
## Problem

The session-admission hardening inside #326 (commit d1118f8) rewrote `web/ui` and silently dropped most of the polished design from 4421944 (`feat(web): polish workbench UI`): the settings dialog (language/theme pickers), the landing logo animation, rich thinking/tool-call rendering, subagent/workflow status cards, the turn rail, the composer activity bar, the favicon, and the engineering-grid backdrop. The merged workbench shipped with a minimal generic transcript instead, far from the workbench experience outlined in #76.

## Value

- Restores the reviewed 4421944 design baseline the web workbench was specified against (#76).
- Re-establishes the UI foundation several open web issues build on: theme/language preferences (#350), structured tool evidence rendering (#345), and capability activity surfaces (#346).
- Keeps every hardening fix from d1118f8 and later (session admission, event recovery, lease/terminal boundaries) — this is a UI restoration, not a revert.

## Approach

- Restored `web/ui/{app.js,index.html,styles.css}` to the 4421944 design, embedded into the current hardened data flow rather than overwriting it: `sessionSwitching`, `applyRuntimeEvent`, prompt admission tokens, and `resetLiveState` are preserved.
- Ported the rendering layer: thinking rows with live timers (timed in `applyRuntimeEvent`, reset in `resetLiveState`), tool lines with icons/argument summaries and right-edge outcome glyphs (✓/✗/running dot), 4+ run grouping, family cards for subagent/workflow calls, per-turn copy/time message actions, turn rail, and composer activity chips adapted to the current `{ items, omitted }` capability projections.
- Restored the settings dialog with browser-local persistence (`openpi.language`, `openpi.web.theme`) matching the original behavior; canonical setup-config integration remains future work tracked by #350.
- Restored `favicon.svg` and its static route in `web/host/web-host.ts`.
- Tests: converted d1118f8-era `doesNotMatch` lock-in assertions back to positive assertions for the revived UI surface; fixtures aligned to current protocol types.

## Validation

- `node --check web/ui/app.js` — OK
- `bun run check` — config contract, biome format/lint, tsc all pass
- `bun run test` — 1236 tests, 1235 pass / 0 fail / 1 skip
- Not run: real-browser manual audit; visual fidelity targets the 4421944 snapshot.

## Impact

[...]

PR #384, opened 2026-09-04
refactor(web): migrate the Web workbench UI to React

6iKUN6:
## Problem

The Web workbench's imperative browser script mixes transcript rendering, Session navigation, streaming state, and interaction logic. This PR replaces it with typed React components and a browser store. Related to #76; the visual reference is #352 at `1b340f2`, while compatibility follows current main `0b282430798939a19412ff83fce277916f5d0cf1`.

The integration review found missing newer-main behavior in the original React branch: exact turn cancellation, canonical theme preference, reliable prompt admission/retry and queue receipts, quiet/stalled connection recovery, and running-state protection. These are restored before merge.

## Value

Give the workbench component, state, and protocol boundaries that are easier to maintain and test, with Vite HMR for development and self-contained production assets. The React implementation carries forward #352's visual work; merging the legacy implementation separately is unnecessary.

## Approach

- React 19/TypeScript, Zustand, Vite, Tailwind, Astryx primitives, sanitized Markdown, and local bundled assets.
- Preserve Pi/OpenPI runtime, Session ownership, authenticated commands, and canonical snapshot/event semantics.
- Port main's canonical light/dark/system preference, exact-turn Stop, native follow-up receipts, stable admission IDs and uncertain retries, heartbeat refresh, and bounded connection reads. Also reject truncated/invalid successful prompt receipts and bound SSE response-header waits.
- Keep same-message streaming pinned when appropriate, force scrolling after send, attribute activity only to the current Session, and preserve background terminal projections and an accessible persistent heading.
- Integrate main's Pi 0.85.1 peer resolution, CLI shutdown fixes, and CI/package isolation. Ship only the fixed Web asset allowlist; build checks reject committed bundle drift.

## Validation

Final integrated source HEAD: `71462796f60246e9e0cf0eb35e3545ba72f8ae99`.

- `bun run check`: passed, including frontend build/typecheck, format and lint.
- `bun run test`: 1,442 Node tests passed, one platform skip; 81 Vitest tests passed.
- `bun run test:web:e2e` using local Chrome: 4 passed, including desktop/mobile, reduced motion, dark running state, exact cancellation request, and axe checks.
- Core regression tests first reproduced failures before repair. Legacy DOM/source-string checks are replaced by store/protocol/component tests against the actual React implementation.
- `npm pack --dry-run --json`: all four production assets included.
- Isolated Pi package discovery: `pi list` identified only this checkout; offline RPC startup exposed the expected OpenPI status/widget without errors.
- Two independent final reviews: clean; the final activity attribution finding was repaired and re-reviewed.
- Final-head remote CI: all passed on Node 22.19.0, Node 24, Node 26, Windows, and Web E2E, including packed CLI and committed-source install smokes: https://github.com/openpi-dev/openpi/actions/runs/34080319155.

Browser tests run the production WebHost. The running/cancellation fixture intercepts canonical snapshots and command receipts; it is not a live-provider or long-running Workflow acceptance claim. The user's installed OpenPI source was not changed.

## Impact

The browser implementation and development/build tooling change. Backend protocol, model-visible tools/context, permissions, Session persistence and canonical configuration do not change. Theme continues to use Pi's existing preference. Production remains local and self-contained, with no Vite/CDN requirement at runtime.

The single JavaScript bundle is approximately 714 kB raw / 213 kB gzip and still triggers Vite's size warning. This is a maintainability migration, without a measured rendering-speedup claim. Historical Markdown parsing during streaming remains a separate performance opportunity.

[Design and integration evidence](https://github.com/6iKUN6/openpi/blob/codex/web-react-mvp/docs/design/OPENPI_WEB_REACT_MVP.md).


6iKUN6:
@tt-a1i Could you please review the technical direction and implementation of this React Web MVP? Thanks.

tt-a1i:
来了












涂少坤
***@***.***


---Original---
From: ***@***.***&gt;
Date: Fri, Sep 4, 2026 17:53 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [openpi-dev/openpi] refactor(web): migrate the Web workbench UI to React (PR #384)


6iKUN6 left a comment (openpi-dev/openpi#384)
 
@tt-a1i 麻烦帮忙 review 一下这个 React Web MVP 的技术方向和实现，谢谢。

scripts/dev-web.mjs
```diff
@@ -1,89 +1,124 @@
...
-  const endpoint = `http://127.0.0.1:${backendPort}/`;
+async function waitForBackend({ backendOrigin, token, startup }) {
+  const endpoint = `${backendOrigin}/api/snapshot`;
   const deadline = Date.now() + 15_000;
   while (Date.now() < deadline) {
+    const startupFailure = startup.getFailure();
+    if (startupFailure) throw startupFailure;
     try {
-      const response = await fetch(endpoint);
+      const response = await fetch(endpoint, {
```
tt-a1i: [P1] Enforce the advertised readiness deadline on each probe. The outer 15-second loop does not bound this awaited `fetch`: a peer can accept the TCP connection and never return headers, leaving `dev:web` stuck here forever and preventing `startup.waitForFailure()` from being observed. Please abort each request at the remaining deadline (and when startup fails), and add a hanging-peer regression test.

web/ui/src/store/web-store.ts
```diff
@@ -0,0 +1,780 @@
...
+          return false;
+        } finally {
+          if (epoch === sessionEpoch && promptAdmissionToken === admission) {
+            promptAdmissionToken = null;
+            set({ promptAdmissionPending: false });
+          }
+        }
+      },
+      setQuery(query) {
+        set({ query: query.trim() });
```
tt-a1i: [P2] Preserve the raw value of this controlled input. Trimming in the setter rewrites `"foo "` to `"foo"` immediately; typing the next character then produces `"foobar"` instead of `"foo bar"`. Keep the raw query in state and trim only the value used for matching, with a test covering spaces in a search term.

web/ui/src/components/Markdown.tsx
```diff
@@ -0,0 +1,52 @@
...
+  } catch {
+    return "";
+  }
+}
+
+export function Markdown({ children }: { children: string }) {
+  return (
+    <div className="markdown">
+      <ReactMarkdown
+        remarkPlugins={[remarkGfm]}
```
tt-a1i: [P2] Preserve the frozen baseline's soft-line-break behavior. PR #352 configured Marked with `breaks: true`, while GFM alone renders `"first line\nsecond line"` without a `<br>`; this component sits under `.message-content`, so it does not inherit the user-message `white-space: pre-wrap` rule and the browser collapses the newline. Add `remark-breaks` (or an equivalent rendering rule) and a parity test.

tests/web/web-store.spec.ts
```diff
@@ -0,0 +1,778 @@
...
+    emit(event: WebEvent) {
+      if (!active) throw new Error("Event stream is not connected");
+      active.onEvent(event);
+    },
+  };
+}
+
+afterEach(() => {
+  sessionStorage.clear();
+  localStorage.clear();
```
tt-a1i: [P2] Keep this suite compatible with the declared `node >=22.19.0` range. Under Node 26.3.0, the native Web Storage global shadows jsdom and unqualified `localStorage` is undefined, so all 21 store tests fail in this cleanup. I reproduced the same failure at this head. Use `window.localStorage` (including the assertion below) or explicitly configure the test environment, then cover a supported newer Node in CI.

package.json
```diff
@@ -58,21 +58,42 @@
     "acorn": "^8.17.0",
     "effect": "^4.0.0-beta.99",
     "jiti": "2.7.0",
-    "marked": "^18.0.5",
     "undici": "8.9.0"
   },
   "devDependencies": {
+    "@astryxdesign/core": "0.5.2",
```
tt-a1i, line 64: [P1] Add notices for the code now shipped in the committed production bundle. `web/dist/app.js` embeds React/ReactDOM, Zustand, Lucide, i18next, eventsource-parser, React Markdown/remark/rehype, Astryx, and related runtime code, but `THIRD_PARTY_NOTICES.md` still lists only the Sessions extension. The relevant MIT/ISC terms and retained copyright notices need to accompany distribution before this can leave draft.

tt-a1i (changes requested):
The React direction is sound and keeps Pi as the runtime source of truth, but this draft is not ready to merge yet. I found two user-visible parity regressions, an unenforced development-startup bound, a supported-Node test failure, and a distribution-notice gap.

I reviewed exact head `13f3032decdcc600e7d46d848b2cc0097fd54d7e`. Locally, `bun run check` and `bun run test:web:e2e` pass (3/3). On Node 26.3.0, the focused Web store suite fails 21/21 because the tests access an unavailable unqualified `localStorage`. The required Node 22 CI job is also currently red: the packed standalone CLI smoke attempts to fetch unavailable `@earendil-works/pi-tui@0.85.0` and receives 404, so the recorded installed-package validation is not current.

Please also update the design record's source boundary to identify this PR and an exact implementation head when the evidence is frozen, rather than only a mutable local branch plus the base and PR #352.

tt-a1i:
Reviewed at exact head `13f3032decdcc600e7d46d848b2cc0097fd54d7e` and submitted [REQUEST_CHANGES](https://github.com/openpi-dev/openpi/pull/384#pullrequestreview-5111852937).

The React/Pi-native direction looks good, but the draft still needs fixes for the unbounded development readiness probe, controlled-search whitespace regression, Markdown soft-break parity, Node 26 Web-store test compatibility, and bundled frontend notices. The required Node 22 packed-CLI smoke is also currently failing with an unavailable `@earendil-works/pi-tui@0.85.0` tarball.

6iKUN6:
@tt-a1i The review feedback has been addressed, the merge conflict is resolved, and CI is green. Could you please re-review when convenient?

PR #358, opened 2026-09-03
feat(web): expose Pi-native turn cancellation

seekskyworld:
## Problem

Implements the first runtime slice of #342. The Web host had no per-turn cancellation endpoint and could only abort sessions during host disposal.

## Value

A browser client can request cancellation without tearing down unrelated Web runtime state, while stale Session requests are rejected.

## Approach

Add an optional `interruptTurn` controller seam backed by Pi `AgentSession.abort()`. Bind the request to the active Session id, return `cancelled` or `already_settled`, emit a cancellation event, and expose `POST /api/prompt/cancel`. Existing test doubles remain compatible through the optional interface member.

## Validation

- `npx tsc --noEmit`
- `node --test --experimental-strip-types tests/web/pi-runtime.test.ts tests/web/web-host.test.ts` (39 passed)
- `git diff --check`

## Impact

- User-visible behavior: adds a server cancellation seam; UI control wiring remains a follow-up.
- Model-visible context/tools: none.
- Runtime/lifecycle: per-turn abort is distinct from host disposal and uses Pi lifecycle.
- Persisted config/data: none.
- Compatibility/risk: old controllers without the optional seam receive a bounded 501 response.

PR #385, opened 2026-09-04
feat(web): add turn-scoped cancellation

### LINKED ISSUE #342: feat(web): support Pi-native turn interrupt and cancellation
## Parent

- #76

## Problem

The current Web workbench can submit prompts and queue follow-up input, but it cannot stop an active model turn. The browser exposes only a Send action; WebRuntimeController has no interrupt/cancel operation, and the runtime only aborts sessions while disposing the whole Web host.

A long, stuck, or accidentally submitted turn therefore cannot be stopped without closing the workbench. Closing the host is not an acceptable per-turn control because it also tears down unrelated Web runtime state.

## Value

- Let users recover immediately from a runaway or unwanted turn.
- Preserve the same cancellation lifecycle and terminal evidence as Pi instead of inventing a browser-only status.
- Establish the command/receipt boundary needed by future capability controls.

## Approach constraints

- Reuse Pi AgentSession cancellation and lifecycle events as the source of truth.
- Bind every cancel request to the expected Web Session and current command/turn epoch.
- Define an exact receipt that distinguishes accepted, already settled, stale session, and failed cancellation.
- Do not infer cancellation from button state or optimistic UI.
- Keep queued follow-up prompt behavior explicit: cancellation must define whether queued input is retained or discarded.
- Preserve controller serialization, SSE resync, shutdown, and multi-client lease boundaries.

## Acceptance criteria

- [ ] While a real provider turn is active, the composer exposes a Stop action.
- [ ] A valid Stop request cancels that turn through Pi native session lifecycle.
- [ ] A stale request cannot cancel a newer turn or a different Session.
- [ ] The UI reaches a canonical cancelled terminal state from runtime evidence.
- [ ] Repeated Stop requests are idempotent and return an explicit receipt.
- [ ] Behaviour for queued follow-up prompts is specified and covered by tests.
- [ ] Refresh/reconnect converges to the same state through snapshot + events.
- [ ] Host disposal remains separate from per-turn cancellation.
- [ ] Focused Web runtime/host/UI tests, bun run check, and bun run test pass.

## Out of scope

- A global pause across every OpenPI capability; that broader problem is tracked in #161.
- Controller handoff between TUI and Web.
- Adding a second cancellation state machine in the browser.

testikun:
Adds a Stop control for the current Pi agent execution without stopping the Web Host. Requests bind to the Session, command, and execution epoch, so stale tabs cannot cancel a later execution.

Closes #342.

The runtime invokes native `AgentSession.abort()` and waits for `agent_settled`. Individual assistant messages, including `toolUse`, do not establish execution completion. Cancellation needs an aborted assistant result; missing terminal evidence remains uncertain. The browser restores the control from snapshots and SSE and does not optimistically claim completion. Queued follow-ups remain under Pi ownership and may continue within the same execution; this does not introduce per-input cancellation or a second queue.

Validation: `bun run check` passed; final integrated focused Web tests passed 75/75. The repair's standalone Node 22 full suite passed 1405 tests with one skip, plus Vitest 30/30. CI is running against the updated head. Earlier live-provider evidence in the issue predates this repair and is not a new runtime acceptance result.


testikun:
CI status note: Node 22/24 completed this PR's repository checks and failed only at the final fresh-install `Smoke-test packed standalone Web CLI` step, where the child exits before the test can read its URL. The same failure now occurs on unrelated PRs #385, #386, and #387.

I reproduced the packed install independently: npm now admits Pi 0.85.0 through OpenPI's published `peerDependencies: "*"`; the CLI then exits with `Cannot find module '@earendil-works/pi-server'`. The checkout/lock-based `bun run check` and full test suite remain green. I recorded the shared packaging regression and suggested peer-range follow-up in #328: https://github.com/openpi-dev/openpi/issues/328#issuecomment-5539313015

No dependency workaround is mixed into this feature PR.

tt-a1i (commented):
## 方向对，有一处应修；请用这份替换 #358

turn 范围（`sessionId+commandId+epoch`）、等 `turn_settled(outcome=cancelled)`、不乐观结算，这些都符合合同。#358 从 `abort()` 返回值推断取消，请关掉那份。

应修：`cancelActiveTurn` 里 `Promise.race([settlement, abortFailure])`。如果 `abort()` resolve 了，但又没有 assistant `message_end`，HTTP 会一直挂着（没有超时）。请给 settlement 加界，或把「abort 成功但未结算」变成明确的非终态/错误，而不是挂起。

CI 红是共享的打包 `openpi web` 冒烟，不是这份 diff 独有的。

tt-a1i (changes requested):
## 需要改：`abort()` 成功但没有 `message_end` 时 HTTP 会一直挂

上一评写成 COMMENT 偏软。`cancelActiveTurn` 里：

```ts
const abortFailure = new Promise<never>((_, reject) => {
  void abortOperation.catch(reject);
});
const terminal = await Promise.race([settlement, abortFailure]);
```

`abort()` resolve、又没有 assistant `message_end` 时，`settlement` 永不 settle，`abortFailure` 也不 reject，HTTP 无限等。这是取消合同上的无界挂起，合入前要加超时或明确的非终态/错误。

方向仍对：`sessionId+commandId+epoch`、等 `turn_settled`、替换 #358。CI 红仍是共享打包冒烟。

testikun:
已按 review 修复并推送到 `c4305ad`：取消请求现在同时等待 settlement、abort 失败和 10 秒有界超时；abort 成功但没有 `message_end/turn_settled` 时会返回明确的 `failed` 错误，不会让 HTTP 请求无限挂起。原有 Pi runtime 测试 22/22 通过，diff 检查通过。

PR #386, opened 2026-09-04
fix(workflows): recover terminal artifact commits

testikun:
## Problem

A Workflow process crash after writing terminal side artifacts but before the final manifest can orphan valid result/transcript evidence. The initial recovery design also allowed an older completed receipt to overwrite newer failed or cleanup facts. Closes #110.

## Value

Verified terminal artifacts become discoverable after a crash without treating arbitrary files as success, and recovery cannot roll back a newer canonical manifest.

## Approach

Publish the terminal reference-free manifest, then a bounded receipt containing its exact SHA-256 predecessor identity, the intended final manifest, and fixed side-artifact names/sizes/digests. Promote only when artifacts validate and the canonical predecessor still matches. Already-committed manifests retain later delivery/resource updates. Missing, modified, malformed or superseding canonical state fails closed. Run recovery at the shared persisted-read boundary and before delivery updates.

## Validation

Integrated current main and repaired the reproduced failed-to-completed rollback. Focused regressions passed 49/49. `bun run check` and standard `bun run test` passed: 1466 Node tests, one platform skip, Vitest 30/30. Two independent final reviews are clean on `3710707286930481c0b95af316abd911d8e6afef`, integrated with main `b0fc5588cd6ae514cc4a64db87ab1c899d2f034f` including #423. Final-head remote CI passed on Node 22.19.0, Node 24 and Windows: https://github.com/openpi-dev/openpi/actions/runs/34078645725. The historical requested change about already-committed identity is addressed, as acknowledged by its reviewer; the final implementation also guards the uncommitted predecessor against newer canonical updates.

## Impact

No model tool, permission or configuration change. Persisted terminal writes briefly include a bounded `.workflow-commit.json` receipt. Legacy runs without receipts remain readable; the unshipped receipt shape now requires predecessorSha256. If an uncommitted manifest changes before recovery, recovery conservatively retains that newer canonical state and the receipt for inspection rather than silently replacing it. The protocol covers process-crash recovery across existing atomic file replacements, not power-loss durability/fsync or a filesystem-wide transaction.

[Architecture and evidence](https://github.com/testikun/openpi/blob/codex/issue-110-artifact-commit/docs/architecture/WORKFLOW_ARTIFACT_COMMIT.md).

PR #387, opened 2026-09-04
feat(web): add bounded transcript content search

testikun:
## Problem

Issue #349 requires transcript-aware server search, but the current Web snapshot exposes only a bounded title/cwd/first-message projection. The open metadata-search PRs do not provide a safe mechanism for scanning Session JSONL content or tool evidence.

Relates to #349. This PR implements the transcript-scanner slice only; it does not duplicate PR #362's endpoint or PR #376's pagination work.

## Value

- Gives the Web host a reusable server-side primitive for finding prior user/assistant text and bounded tool evidence.
- Keeps Pi Session files authoritative while making partial, malformed, changed, unavailable, and budget-limited outcomes observable.
- Preserves exact Session/message/turn/workspace/source/timestamp provenance without retaining an unbounded index.

## Approach

- Add `searchWebTranscripts()` over a caller-authorized Session catalog; the scanner never discovers or widens filesystem authority.
- Enforce hard caps for catalog files, total bytes, per-file bytes, line bytes, retained results, snippet bytes, and wall time; support `AbortSignal` cancellation.
- Require regular non-symlink files within canonical allowed roots and exact workspace membership, then validate the Session header identity before publishing matches.
- Scan JSONL incrementally in fixed-size chunks, retain only a bounded newest-result set, propagate turn identity across intervening Pi entries, and re-check file identity after scanning.
- Search user/assistant text and tool names only; exclude tool-call arguments/input and tool-result bodies/details. Return control-safe, bidi-safe, UTF-8-bounded plain-text snippets with explicit truncation evidence.
- Fail soft per file for malformed, changing, deleted, oversized, and unauthorized inputs, with exact `partialReasons` and accounting.

The existing lightweight metadata filtering remains separate. Endpoint composition, pagination/cursors, and derived index invalidation remain follow-up slices of #349.

## Validation

Integrated with main `c1c60cd9e357044b53042ab23963d053b46dae9f`, preserving the author's commits. The scanner and its test are byte-identical to reviewed head `c2f395b`; no tool registration, prompt context, or host route is added.

- `bun run check`: passed with the frozen lockfile and Pi 0.85.1.
- `bun run test` on integrated tree `e5af1744e3525d5859929eff2c44c076d5a3f357`: Node 1436 passed, 1 skipped, 0 failed; Vitest 30 passed.
- Transcript-search tests: 9/9 passed, included in the full suite.
- Independent probes confirmed tool arguments/input/results/details do not enter matches or snippets; tool-name searches return only the name.
- Earlier member-CLI runs failed on two plan-mode rendering assertions and a setup child timeout. The same failures reproduced on unmodified main; an unchanged-tree serial run from the main review environment passed. The exact environmental trigger was not established; no production or environment workaround was applied.
- No manual UI or installed-Pi acceptance is claimed. The scanner is not yet connected to a host endpoint or UI.

## Impact

- User-visible behavior: None until a Web endpoint/UI composes the scanner.
- Model-visible context/tools: None.
- Runtime/lifecycle: Adds a read-only, cancellable, bounded Session-file scanner; no second Session store or index.
- Persisted config/data: None.
- Compatibility/risk: Additive module only. The caller must supply the authorized Session catalog and roots; the scanner fails closed on provenance or path mismatches.


testikun:
CI status note: Node 22/24 completed this PR's repository checks and failed only at the final fresh-install `Smoke-test packed standalone Web CLI` step, where the child exits before the test can read its URL. The same failure now occurs on unrelated PRs #385, #386, and #387.

I reproduced the packed install independently: npm now admits Pi 0.85.0 through OpenPI's published `peerDependencies: "*"`; the CLI then exits with `Cannot find module '@earendil-works/pi-server'`. The checkout/lock-based `bun run check` and full test suite remain green. I recorded the shared packaging regression and suggested peer-range follow-up in #328: https://github.com/openpi-dev/openpi/issues/328#issuecomment-5539313015

No dependency workaround is mixed into this feature PR.

tt-a1i (changes requested):
## 需要改：tool-call 参数会进搜索 snippet

`web/search/transcript-search.ts` 的 `messageEvidence` 对 `toolCall` 做了：

```ts
const payload = tool.arguments === undefined ? tool.input : tool.arguments;
text: `${tool.name} ${JSON.stringify(payload)}`
```

PR 正文写排除 private result details，但 **arguments / input 整段可搜、可进 snippet**。工具参数里的 token、路径、密钥会漏到 Web 搜索面。

`toolResult` 正文也会整段进 index。请把 arguments 从 searchable text 拿掉，或只留 allowlisted 的非敏感字段，并补一条「密钥不得出现在 snippet」的测试。

原语本身（文件/大小/命中上界、symlink 检查、AbortSignal）是好的，泄漏修完再合。现在还没有 HTTP endpoint，但这个函数一旦接上就会爆。

testikun:
已按 review 修复并推送到 `c2f395b`：toolCall 的 arguments/input 和 toolResult 的正文不再进入搜索 evidence；工具名仍保留为非敏感 provenance。新增回归测试验证参数中的 Bearer token 和结果中的 apiKey 不会出现在搜索命中或 snippet 中。transcript-search 测试 9/9 通过，Biome 和 diff 检查通过。

tt-a1i (approved):
Exact-head review of integrated 3e04d856254190ad4097aadadef2872555236e0d (tree e5af1744e3525d5859929eff2c44c076d5a3f357). The previous tool-payload indexing blocker is fixed: tool calls/results contribute names only; arguments, input, result bodies and details are excluded. Focused tests and independent probes exercised those exclusions along with path, provenance and resource bounds. No blocking finding was confirmed.

The main integration is clean and preserves the reviewed scanner/test bytes; the delta is only those two new files. No tool registration, model context, persistence writes or host endpoint is changed.

Validation on the integrated tree: bun run check passed; serial bun run test passed (Node 1436 passed, 1 skipped, 0 failed; Vitest 30 passed), including 9 transcript-search tests. Earlier member-CLI full runs had two rendering failures and a setup timeout that also reproduced on pure main; those logs are retained, the exact environment trigger remains unknown, and no workaround was applied to obtain this pass.

This is approval of the scanner primitive, not end-to-end search UI or installed runtime acceptance. Required remote CI and repository approval rules still apply; no merge was performed.

web/search/transcript-search.ts
```diff
@@ -0,0 +1,648 @@
...
+  tokens: string[],
+  byteBudget: number,
+  limits: WebTranscriptSearchLimits,
+  signal: AbortSignal | undefined,
+  expired: () => boolean,
+): Promise<FileScanResult | FileScanFailure> {
+  let handle: Awaited<ReturnType<typeof open>> | undefined;
+  let bytesConsumed = 0;
+  try {
+    const beforePath = await stat(canonicalPath);
```
tt-a1i, line 376: [P2] Bind the opened file to the identity that passed authorization

The earlier lstat/realpath validation is discarded before this fresh stat/open sequence. If the validated allowed/session.jsonl is replaced by a symlink to outside.jsonl before this stat, both stat and open follow the outside target, and the subsequent identity checks agree with that new target. Reproduced using real files and a scheduled unlink+symlink in that window: an outside file with matching Session header produces `needle outside evidence` under the authorized path with partial=false. This violates the explicit non-symlink/allowed-root boundary. Please validate the opened descriptor against the previously authorized identity and reject path substitutions, with a race regression test.

tt-a1i (changes requested):
Follow-up exact-head review found a new P2 not covered by the prior approval: authorization/path validation is not bound to the file descriptor actually scanned. The scanner still has value and the earlier tool-payload exclusion remains fixed, but this new boundary failure blocks merge. Focused tests passed 9/9; a deterministic real-filesystem race probe replaced the validated file with an outside-root symlink before scanSessionFile stat/open and returned outside transcript content with partial=false. The probe instruments only scheduling via the fs.stat seam; it does not stub file contents or scan results. No source changes or merge performed.

testikun:
Fixed and pushed as `da85f3d` (includes the earlier scan binding fix and latest `main`).

- Transcript scanning now validates the opened descriptor against the identity authorized by the pre-scan `lstat`/`realpath`, and verifies the path/descriptor again after reading.
- Path substitution or symlink replacement is reported as a changed session and cannot expose outside-file evidence.
- Validation: Node 24 transcript-search tests passed (9/9).

tt-a1i (changes requested):
Reviewed exact head da85f3d37c20efbaee4a40feaa7e188368d05706.

This is a valuable bounded, read-only transcript scanner, and the earlier tool-payload leak and pathname/FD identity issues are substantially improved. One P1 authorization/provenance gap remains:

- [P1] The scanner verifies the Session header id but not its cwd. It authorizes and labels results using catalog cwd, so a stale/substituted catalog entry can point to a transcript with the same id but an unauthorized header workspace. I reproduced an outside-workspace header returning one match labeled as the allowed workspace with partial=false. Require the opened Session header cwd, canonicalized under the same policy, to equal the authorized catalog workspace before retaining any evidence.

[P2] The new validation-to-open substitution fix also lacks a deterministic regression test; current symlink coverage only rejects a symlink that already exists before opening.

Focused tests pass 9/9 and all hosted checks are green. No endpoint, tool, prompt, persistence, or Session mutation is added, but the P1 crosses the core workspace privacy boundary and blocks merge.

testikun:
已推送 `972f542`，并同步当前 main。针对本次评审的两个问题：

- P1（header cwd）：现在对允许的 workspace、catalog cwd 和实际打开的 Session header cwd 使用相同的 `realpath` 规范化规则；header 必须包含绝对 cwd，且规范路径必须等于已授权 catalog workspace。缺失、相对路径或跨 workspace 的 header 都不会返回 evidence，并报告 partial/malformed-session。补充了同 id、不同 workspace 的回归，以及合法路径别名的覆盖。
- P2（验证到 open 的替换）：新增真实文件系统的确定性竞态测试，在授权完成后、打开文件前替换成外部 symlink，确认结果为空且报告 changed-session；测试只控制调度，不伪造文件内容或扫描结果。

验证：transcript-search 11/11 通过，`bun run check` 和 `bun run test` 通过（Vitest 220/220）。完整测试使用 `PI_HYPERLINKS=0`，避免本地 Ghostty 的 OSC 8 输出影响现有展示断言。未执行安装态 Pi 或手工 UI 验收；此模块仍未接入 endpoint/UI。

testikun:
已核对当前 head 972f542。评审 P1 已由现有提交解决：Session header 的绝对 cwd 与 catalog workspace、allowed workspace 均按同一 realpath 规则规范化并严格相等；缺失、相对或跨 workspace header 只产生 partial/malformed 结果，不保留 evidence。验证到 open 的 symlink 替换也有确定性真实文件系统回归。验证：transcript-search 11/11 通过。当前没有未解决问题。

tt-a1i (commented):
审查提交：`972f5426d8580c60a32d5802911f565334cfe8a7`。

### 验证范围

Reviewed entire scanner and all tests. Exact-head 11 tests passed, including real Pi persisted Session, tool-payload exclusion, canonical header workspace, validation-to-open symlink substitution, limits and cancellation. Current implementation explicitly compares the opened file against authorizedIdentity, so the older review text claiming it discards that identity does not describe these bytes.

### 限制与后续

Scanner primitive only: no production caller, HTTP endpoint or UI. Time budget is cooperative around filesystem awaits, not a hard I/O deadline. Catalog is capped before sorting/filtering; does not promise globally newest results. No end-to-end installed search acceptance.

PR #388, opened 2026-09-04
feat(web): add attachment staging lifecycle

testikun:
## Problem

The attachment work tracked by #344 has admission metadata validation in review, but no host-owned lifecycle for canonical temporary bytes. Browser filenames must not become filesystem authority, and staged bytes need exact command ownership and deterministic cleanup.

## Value

Future composer and multimodal prompt work gets a fail-closed storage primitive with bounded resource use, single-consume semantics, and explicit stale or settled receipts.

## Approach

Add a private attachment staging store that:

- writes payloads under server-generated file identities in a private host directory;
- binds every batch to the exact canonical workspace, Session id, and command id;
- serializes stage, consume, discard, and disposal mutations;
- enforces injected count, per-file, per-command, store-wide byte, and settled-receipt limits;
- rejects changed files and symlinks before reading;
- removes staged artifacts on consume, discard, failed integrity checks, and store disposal.

This is a focused backend slice and intentionally does not duplicate #365/#380 metadata admission validation or #384 UI work.

## Validation

- bun run check
- bun run test: 1245 passed, 0 failed, 1 skipped; Vitest 30 passed
- focused attachment staging tests: 4 passed
- git diff --check

## Impact

- User-visible behavior: none until the composer integration lands.
- Runtime/lifecycle: adds an opt-in host-owned temporary byte lifecycle; no current request path changes.
- Persisted config/data: none; artifacts are private and temporary.
- Compatibility/risk: additive module only. Limits are injected so the final upload path can use the authoritative admission policy without a second default.

Related to #344.


testikun:
CI context: both Linux Node jobs reach the pre-existing packed standalone Web CLI smoke, install the tarball, and then the server exits before announcing its origin (`kill: No such process`). Windows and the other checks pass. This matches the independently reproduced fresh-install peer-dependency failure tracked in #328, where npm selects Pi 0.85.0 and `@earendil-works/pi-server` cannot be resolved. Evidence: https://github.com/openpi-dev/openpi/issues/328#issuecomment-5539313015

This PR does not change package metadata, CLI startup, or peer-dependency resolution. Its local check, focused tests, full Node suite, and Vitest suite pass as recorded in the PR body.

tt-a1i (changes requested):
## 需要改：暂存没有 TTL，name/mime 也没有自己的上界

`web/runtime/attachment-staging.ts`：

- 崩溃且没走到 `dispose()` 会留下 `0o700` 目录，没有 TTL / 启动回收。
- `validatePayloads` 不限制 `name` / `mime`，完全指望还没合的 #365/#380。

staging 不该自带一份政策，但也不该在准入 PR 落地前把无界元数据写进 store。请先合 #365+#380（或把同样的边界搬进来），并给未 `dispose` 的根目录加 TTL 或启动清理。

绑定、独占队列、symlink/size fail-closed 这些是对的。

testikun:
已按 review 修复并推送到 `9ebd185`：

- staging store 启动时会按 `stagingTtlMs`（默认 1 小时）回收父目录中过期的 `.openpi-web-attachments-*` 根目录，崩溃后未走 `dispose()` 的目录不会永久遗留。
- `name` 和 `mime` 各自限制为最多 256 UTF-8 bytes，并在准入阶段 fail-closed。
- 新增过期根目录回收和元数据上界回归测试。

attachment-staging 测试 5/5 通过，Biome 和 diff 检查通过。

web/runtime/attachment-staging.ts
```diff
@@ -0,0 +1,441 @@
...
+    for (const entry of await readdir(parent)) {
+      if (!entry.startsWith(".openpi-web-attachments-")) continue;
+      const candidate = join(parent, entry);
+      try {
+        const candidateStat = await lstat(candidate);
+        if (
+          candidateStat.isDirectory() &&
+          now - candidateStat.mtimeMs > ttl
+        ) {
+          await rm(candidate, { recursive: true, force: true });
```
tt-a1i, line 242: [P1] Do not reclaim a live store based only on its directory mtime

`stagingTtlMs` is documented as the age for reclaiming an abandoned store, but this loop has no evidence that the owner is stopped. Two stores may use the same parent; `create` does not require exclusive ownership of that parent. If store A has an unconsumed batch and its root mtime is older than the TTL, creating store B recursively deletes A even though A was never disposed. A subsequent `consume` fails with `staged attachment integrity check failed`, and the staged bytes are gone.

Reproduction on this head: create A under a unique temporary parent; stage a batch; use `utimes` to age only A's root beyond the configured TTL (simulating an idle live store); create B under the same parent; `lstat(A.root)` returns ENOENT and A.consume returns failed. The probe exits 0 after asserting this loss.

Reclaim only roots whose ownership/liveness is proven abandoned, or enforce an ownership boundary that makes cross-store deletion impossible. Add a regression retaining A's unconsumed batch across B.create while still testing genuine orphan cleanup. A timestamp alone does not prove abandonment.

tt-a1i (changes requested):
Exact-head re-review of 9ebd18599ad50d73e43847284fd0a1ea9a303a69. The previous metadata bounds and missing cleanup findings have been addressed; focused tests pass (5/5). However, the new startup reclamation can delete a different live store and its unconsumed attachments. This is a lifecycle/data-loss blocker, independently reproduced with the real module in a unique temporary parent. No host/UI or deployed impact is claimed; the store is not yet wired to an upload route.

testikun:
已按最新 P1 review 修复并推送 `f8ac1bd`：启动回收不再仅凭目录 mtime 删除可能仍存活的 store。每个新 store 在父目录创建 PID owner marker；回收时若 marker 对应进程仍存活则保留，旧无 marker 根目录仍按 TTL 回收。marker 位于 store 外部，不会进入 attachment payload 目录。验证：attachment-staging 测试 5/5 通过，git diff --check 通过。

testikun:
已按该阻塞意见修复：启动回收现在使用 sibling owner marker，并仅在 owner PID 不存活时回收；活动进程的 staging store 会保留。focused attachment-staging tests 5/5 通过。请基于最新提交复审。

testikun:

testikun:

web/runtime/attachment-staging.ts
testikun: @tt-a1i 已修复：每个 store 有 sibling PID owner marker；回收时 owner 存活则保留，旧无 marker 根目录仍按 TTL 回收。live-store 与 orphan 回归均通过。

testikun:
@tt-a1i 已补齐 TTL、name/mime 256 UTF-8 bytes 限制及回归测试；随后又补上 owner/liveness proof，避免启动回收删除仍存活的 store。

testikun:
@tt-a1i 已修复 P1：startup reclaim 现在读取 sibling PID owner marker；owner 存活则保留，旧无 marker 根目录才按 TTL 回收。live-store/orphan 回归通过。

web/runtime/attachment-staging.ts
```diff
@@ -0,0 +1,471 @@
...
+    if (this.closing) {
+      return Promise.reject(
+        new WebAttachmentStagingError("STORE_CLOSED", "attachment store is closed"),
+      );
+    }
+    return this.exclusive(async () => {
+      this.assertOpen();
+      validateBinding(binding);
+      const totalBytes = this.validatePayloads(payloads);
+      if (this.stagedBytes + totalBytes > this.limits.maxStagedBytes) {
```
tt-a1i, line 279: [P2] Bound active stages independently of payload bytes

Zero-byte payloads are accepted by validatePayloads, so this byte-only store admission never advances for them. Each successful stage still creates a directory/file and retains a Map record, while maxSettledReceipts only bounds settled records. On this exact head, a store configured with maxAttachments=1, maxAttachmentBytes=1, maxTotalBytes=1, maxStagedBytes=1, maxSettledReceipts=1 accepted ten separate zero-byte batches and retained ten directories; repetition has no configured bound. Add an active-stage/count or charged-overhead bound (or reject empty payloads if that is the intended contract), with a regression proving repeated submissions eventually reject without leaking partial staging artifacts.

web/runtime/attachment-staging.ts
```diff
@@ -0,0 +1,471 @@
...
+      this.trimSettledReceipts();
+      return { status: "discarded" };
+    });
+  }
+
+  dispose() {
+    if (this.disposePromise) return this.disposePromise;
+    this.closing = true;
+    this.disposePromise = this.exclusive(async () => {
+      await rm(this.directory, { recursive: true, force: true });
```
tt-a1i, line 402: [P2] Remove the sibling owner marker during normal disposal

create writes `${directory}.owner` beside the root, but dispose only removes the root. Reproduction on this head: create a store under a fresh temporary parent, stage a batch, await dispose, and readdir(parent) still returns `.openpi-web-attachments-<id>.owner`. Startup reclamation only removes markers together with matching directories, so orphan marker files are never reclaimed and accumulate across normal create/dispose cycles, increasing future startup scans. Dispose the owned marker as part of the same cleanup lifecycle and test that a fresh dedicated parent is empty after repeated create/dispose cycles.

tt-a1i (changes requested):
Exact-head review of be9eab7c2732df8d2a40254a35e8f382bcded973. The staging design has useful binding and cleanup boundaries, but two independently reproduced P2 resource-lifecycle problems remain. Please fix these in this PR and add regression coverage; no author code was modified during review. This module is not yet connected to an upload route, so these are component findings rather than claims about deployed behavior. Related tracking: #441.

testikun:
已推送 `51706c6`（同步最新 main）。本次处理评审中的两个未解决问题：新增显式 `maxStagedBatches` 活跃批次数上限，防止零字节批次无限增长；dispose 同时删除 owner marker，并用活跃 store 注册表防止同进程创建新 store 时回收仍在使用的旧 store。新增零字节批次、存活 store TTL、marker 清理回归。专项 attachment-staging 测试 7/7 通过。

tt-a1i (changes requested):
Reviewed exact head 51706c6912e17e4c32c46c91b67099a5dd742a51.

The host-owned staging primitive is useful and is not wired into current user/model behavior, but two reproducible P1 integrity/lifecycle blockers remain:

- [P1] Changed bytes can be accepted as canonical. The store records and verifies path/type/size, but not content identity. On the exact head, staging safe, overwriting the payload with same-length evil, then consuming returns consumed with evil. Persist and verify a digest or authoritative opened-file identity, and close the check/read TOCTOU.
- [P1] discard marks the batch settled before deletion succeeds. If rm fails, the first call rejects, but a retry reports settled/discarded while the bytes remain. Only settle after cleanup succeeds, or persist an explicit failed/retryable cleanup state.

[P2] Workspace ownership is also raw absolute-string equality rather than a proven canonical path, so aliases can become distinct owners.

Focused tests and required CI are green, but they do not cover these exact failure cases. The P1s block merging this as the canonical attachment-byte authority.

testikun:
已核对当前 head 5b921e1。评审 P1 已由现有提交解决：staging 时持久化 SHA-256，consume 在 O_NOFOLLOW 已打开描述符上校验 inode/size、读取后再次校验 size 和 digest，因而同长度内容替换或打开时替换均 fail closed；discard 仅在 cleanup 成功后标记 settled，失败保持可重试。验证：attachment-staging 10/10 通过。当前没有未解决问题。

tt-a1i (changes requested):
审查提交：`5b921e1034749aecd12633633bf8c8e86f303ba6`。

需要修改后再合并。

### 具体问题

**P2** · [web/runtime/attachment-staging.ts:425](https://github.com/openpi-dev/openpi/blob/5b921e1034749aecd12633633bf8c8e86f303ba6/web/runtime/attachment-staging.ts#L425)

dispose caches a rejected cleanup promise permanently. A transient rm failure leaves staged private files plus owner marker, but subsequent dispose calls return the same rejection without retrying; ACTIVE_STORE_DIRECTORIES/live PID also prevents startup reclamation while the process lives. Retain closing admission but allow cleanup retry after a failed disposal.

### 验证范围

Full store and tests reviewed; focused suite passed. Injected one rm failure during disposal, restored real fs.rm, then retried: samePromise=true, cleanupAttempts=1, private root and marker still present.

### 限制与后续

Component not wired to an upload route. Old same-length tampering, open substitution, discard retry, canonical-workspace and active-root TTL findings addressed in latest head. No deployed/user upload claim.

PR #361, opened 2026-09-03
feat(web): add session unarchive operation

seekskyworld:
## Problem

Implements the reversible archive-management slice of #347. The Web API could archive sessions but offered no way to restore one.

## Value

Archived Session metadata can be restored without deleting or rewriting the canonical Session file.

## Approach

Add `PiWebAdapter.unarchiveSession` using the existing atomic archive metadata transaction and expose `POST /api/sessions/unarchive`. The operation validates the canonical Session through the same adapter boundary and emits a `session_unarchived` event.

## Validation

- `npx tsc --noEmit`
- `node --test --experimental-strip-types tests/web/web-host.test.ts tests/web/pi-adapter.test.ts` (32 passed)
- `git diff --check`

## Impact

- User-visible behavior: restores an archived Session through the API.
- Model-visible context/tools: none.
- Runtime/lifecycle: no Session replay or side effects.
- Persisted config/data: updates existing `archived-sessions.json` atomically.
- Compatibility/risk: archive view/UI action remains a follow-up; existing archive behavior is unchanged.

Issue #347, opened 2026-09-02
feat(web): complete archive, history, recovery, and fork session lifecycle


sowen1023:
## Parent

- #76

## Problem

The Web workbench has a deliberately separate Web Session directory and a partial Session UI. It can create, select, rename, and archive Web Sessions, but archived entries disappear with no archive view or unarchive action. There is no reviewed Session deletion flow, no read-only browser for compatible terminal Pi history, and no real fork/branch operation.

Historical message editing or regeneration must not be simulated by changing browser labels or resending text while presenting it as an in-place mutation.

## Value

- Let users recover archived work and manage Session retention safely.
- Make existing Pi history discoverable without merging authority or storage.
- Establish honest branch/fork semantics for iterative work.

## Approach constraints

- Pi Session files and lifecycle remain authoritative; Web indexes are derived.
- Keep Web-owned Sessions distinguishable from terminal/default Pi Sessions.
- Historical terminal Sessions are observer-only until an explicit, serialized control-transfer design exists.
- Deleting sidebar metadata, archiving a Session, and deleting persisted Session data are distinct operations.
- Destructive deletion requires native reviewed confirmation and an exact canonical target.
- Resend-as-new-message must remain visibly different from edit, regenerate, branch, or fork.
- Fork is implemented only against accepted Pi Session semantics with parent provenance and side-effect boundaries.

## Acceptance criteria

- [ ] The sidebar has an archived view with bounded pagination/search.
- [ ] Users can unarchive a Session and recover it in its original workspace grouping when valid.
- [ ] Session deletion shows the exact persisted target, uses native confirmation, and has explicit success/failure evidence.
- [ ] Compatible terminal/default Pi Sessions can be discovered and inspected read-only without copying them into Web storage.
- [ ] Session origin, storage source, active controller, and read-only state are visible.
- [ ] Resend-as-new-message is never presented as historical editing.
- [ ] If fork is supported, the new Session records parent Session/turn provenance and does not replay prior side effects.
- [ ] Unsupported edit/regenerate/fork actions remain absent rather than faked.
- [ ] Archive, unarchive, delete, observer history, restart, and corrupt/missing Session cases are tested.
- [ ] Bounded indexing and transcript projection remain enforced.
- [ ] README documents Web-vs-terminal Session behavior.
- [ ] bun run check and bun run test pass.

## Related work

- #193 defines broader Session lifecycle, recovery, replay, and side-effect boundaries.
- #166 researches observer/controller sharing.

[...]

PR #391, opened 2026-09-04
feat(web): list archived sessions

testikun:
## Problem

The Web archive lifecycle tracked by #347 can hide Session entries, and #361 adds an unarchive mutation, but there is no bounded read API that can discover archived Sessions for an archive view.

## Value

Archived conversations become recoverable through a stable, searchable projection without copying or rewriting canonical Pi Session data.

## Approach

- Reuse Pi SessionManager metadata and the existing archived-sessions.json set as the only sources of truth.
- Add a read-only archived Session projection with a 50-item page cap, 5,000-record scan cap, and bounded normalized query.
- Use opaque cursors bound to both the last Session identity and normalized query.
- Return explicit invalid-query and stale-cursor receipts.
- Report matching rows omitted by the page and source records not scanned by the hard bound.
- Expose authenticated GET /api/sessions/archived.

This is distinct from #361 unarchive mutation work. It does not add deletion, fork, historical editing, or UI.

## Validation

- bun run check
- bun run test: 1242 passed, 0 failed, 1 skipped; Vitest 30 passed
- focused adapter and Web Host tests: 33 passed
- git diff --check

## Impact

- User-visible behavior: authenticated API consumers can browse archived Session metadata; no UI changes.
- Runtime/lifecycle: read-only bounded indexing over Pi-owned Session metadata.
- Compatibility/risk: additive endpoint and adapter method; stale cursors fail closed rather than silently skipping entries.

Related to #347.


testikun:
CI context: both Linux Node jobs reach the pre-existing packed standalone Web CLI smoke, install the tarball, and then the server exits before announcing its origin (`kill: No such process`). Windows and the other checks pass. This matches the independently reproduced fresh-install peer-dependency failure tracked in #328, where npm selects Pi 0.85.0 and `@earendil-works/pi-server` cannot be resolved. Evidence: https://github.com/openpi-dev/openpi/issues/328#issuecomment-5539313015

This PR does not change package metadata, CLI startup, or peer-dependency resolution. Its local check, focused tests, full Node suite, and Vitest suite pass as recorded in the PR body.

testikun:

testikun:

web/adapter/pi-adapter.ts
```diff
@@ -284,6 +329,102 @@ export class PiWebAdapter {
...
+    });
+    let start = 0;
+    if (options.cursor !== undefined) {
+      const cursor = decodeArchivedSessionCursor(options.cursor);
+      if (!cursor) return { status: "invalid" as const };
+      if (cursor.query !== normalizedQuery) {
+        return { status: "stale_cursor" as const };
+      }
+      const cursorIndex = matches.findIndex(
+        (session) => session.id === cursor.sessionId,
```
tt-a1i: [P2] Bind the cursor to a unique Session file identity

Using only session.id cannot distinguish copied Session files that retain their original ID. With three archived files ordered newest/middle/oldest, where the first two share an ID and limit=1, actual requests return newest -> middle -> middle -> middle indefinitely: findIndex keeps resolving the first matching ID, and the oldest entry is unreachable. Bind the cursor to the file identity (and validate it against the same result set) rather than ID alone. Add a real-file duplicate-ID pagination regression proving each file is visited once and pagination terminates.

web/adapter/pi-adapter.ts
```diff
@@ -284,6 +329,102 @@ export class PiWebAdapter {
...
+    }));
+    const pageEnd = start + sessions.length;
+    const hasMoreMatches = pageEnd < matches.length;
+    const recordsUnscanned = Math.max(0, allSessions.length - scanned.length);
+    return {
+      status: "ok" as const,
+      sessions,
+      ...(hasMoreMatches && sessions.length > 0
+        ? {
+            nextCursor: encodeArchivedSessionCursor({
```
tt-a1i, line 418: [P2] Ensure every generated cursor is valid for the next request

Valid queries can produce cursors that the same endpoint rejects. Real-file reproduction: two matching archived Sessions, query="中".repeat(160), limit=1 gives status ok and a 740-character nextCursor, but replaying it returns invalid/HTTP400 because WEB_MAX_ARCHIVED_SESSION_CURSOR is512. Separately query="ﬃ".repeat(60) has raw length60 but NFKC length180; the generated cursor is only296 characters, yet decoding rejects its normalized query against the160-character raw-query cap. Keep input normalization, encoding and decoder limits consistent, or use a bounded query fingerprint, and test replay of the returned cursor for Chinese and NFKC-expanding queries.

tt-a1i (changes requested):
The archived-history API has useful value, but two P2 pagination failures were reproduced using real Pi Session files on this exact head. Please address the inline findings and add regression tests before merge. No implementation changes were made during review. Existing focused tests pass 41/41; they do not cover these cases.

testikun:
Fixed and pushed as `ee70b28` (merged latest `main`).

- Archived cursors now bind to the exact Session file path instead of only the Session ID, so copied/duplicate IDs paginate without looping.
- Cursor query state uses a bounded SHA-256 fingerprint of the normalized NFKC query, keeping generated cursors replayable under the 512-byte limit.
- Added duplicate-ID, long CJK, and NFKC-expanding query regressions.
- Validation: Node 24 archived adapter regression passed; `bun run check` passed after formatting.

tt-a1i (approved):
Approved exact head ee70b284d919a598c74668b7181804eeec851bde.

This is a valuable read-only archived-Session projection with no model-context, mutation, execution-authority, or Session-lifecycle change. Exact-head focused tests pass 44/44, required CI is green, and the current-main merge-tree is clean.

Remaining P2/P3 follow-ups are non-blocking under the merge policy: generated cursors can be unreplayable for very long paths; activity-order changes can silently skip an item; the projection omits source/origin/controller/readOnly; the 5,000 cap is after Pi listAll I/O; and the route is not GET-only. Please track these as follow-up correctness/acceptance work.

PR #409, opened 2026-09-05
fix(tui): avoid stale Windows autocomplete rows

### LINKED ISSUE #407: fix(tui): prevent stale autocomplete rows on Windows terminals
## Summary

On Windows, the default regular/main-screen TUI can leave previous slash-command autocomplete rows on screen while typing. In openpi this looks like `login`/`logout` (and the `/`, `/l`, `/lo` intermediate states) are registered multiple times, but the command registry is not duplicated; the stale rows are terminal redraw artifacts.

## Reproduction

Environment observed:

- Windows terminal (PowerShell/Windows Terminal)
- Pi 0.85.0
- `@tt-a1i/openpi` 0.5.0
- openpi extensions loaded from a single package entry

Steps:

1. Start openpi in the default `regular` TUI mode.
2. Type `/` and then `/lo` without submitting.
3. Observe that previous autocomplete rows remain visible and overlap the current list.

Expected: only the current autocomplete list is visible.

Actual: earlier rendered rows remain visible, making commands appear duplicated. The command still resolves correctly.

## Root cause

The regular TUI uses differential rendering against the main terminal screen. On Windows, the redraw/cursor bookkeeping can fail for the multi-line autocomplete editor, leaving stale rows in the terminal scrollback. The same class of symptom is tracked upstream in [earendil-works/pi#6300](https://github.com/earendil-works/pi/issues/6300).

## Proposed fix

Add an openpi Windows terminal compatibility layer that prevents this class of stale redraws, while preserving an explicitly configured user choice. The safest fallback is fullscreen/alternate-screen mode; a regular-mode forced redraw fallback may also be considered if the extension API supports it without replacing user UI components.

The fix should include:

- Windows-only behavior;
- no change when the user explicitly chooses a TUI mode;
- a regression test for the compatibility decision;
- setup documentation explaining how to override the behavior.

This issue is intended to track the openpi-side mitigation and the follow-up pull request.

zhouyuanxinand:
## Problem

Fixes #407. On Windows, Pi's regular (main-screen) TUI renderer can leave stale autocomplete rows on screen during differential redraw, making slash commands appear duplicated even though the command registry is correct.

## Value

The compatibility extension removes the stale-row symptom for affected interactive Windows sessions while preserving Pi's existing TUI and terminal preferences. It does not alter RPC, print, JSON, or non-Windows sessions.

## Approach

- Install the compatibility widget only for interactive Windows TUI sessions.
- Read the raw global and project `terminal.clearOnShrink` settings. If neither scope explicitly configures the setting, enable Pi's supported clear-on-shrink behavior on the current regular renderer only.
- Preserve an explicit `terminal.clearOnShrink` value, including `false`, and leave the selected `tuiMode` unchanged. Settings read errors fail closed.
- Do not call Pi settings setters, flush persistence, or emit a restart/success notification. Users who want a persistent mode or terminal preference can use Pi's native `/settings`, `settings.json`, or `--tui-mode` flow.
- Clean up the widget on session shutdown and keep fullscreen rendering unchanged.

## Validation

- Required PR CI on commit `4465e9d`: Node 22.19.0, Node 24, and Background terminals (Windows) — all passed.
- Native Windows visual acceptance was not run in this environment; the change is limited to the existing renderer hook and has no visual asset changes.

## Impact

- User-visible behavior: affected Windows regular TUI sessions get session-local stale-row cleanup when no explicit clear-on-shrink preference exists; no restart is required and no success notice is shown.
- Persisted config/data: read-only inspection only; no global or project settings are written, and explicit values are preserved.
- Runtime/lifecycle: limited to interactive Windows TUI sessions; the compatibility widget is removed on shutdown.
- Compatibility/risk: fullscreen, RPC, print, JSON, non-Windows, and explicitly configured terminal preferences remain unchanged.


extensions/windows-tui-compatibility/index.ts
```diff
@@ -0,0 +1,133 @@
...
+          globalTuiMode: globalSettings.tuiMode,
+          projectTuiMode: projectSettings.tuiMode,
+          explicitCliTuiMode,
+        }) ||
+        settingsManager.setTuiMode === undefined
+      ) {
+        return;
+      }
+
+      settingsManager.setTuiMode("fullscreen");
```
tt-a1i: [P2] Keep the Windows workaround from silently changing global Pi preferences

Every interactive Windows session without an explicit tuiMode reaches this write; there is no detection of stale rows. Pi SettingsManager.setTuiMode persists a global setting, so opening OpenPI in one workspace also changes subsequent Pi sessions in other workspaces. I reproduced this with the locked real SettingsManager and two temporary workspace paths sharing one agent directory. This exceeds a session-local rendering workaround and conflicts with the repository requirement to preserve Pi-owned preferences and side-effect-safe installation. Please make the persistent mode change an explicit user choice through the native Pi settings flow, or keep the mitigation scoped to the affected session.

extensions/windows-tui-compatibility/index.ts
```diff
@@ -0,0 +1,133 @@
...
+      await settingsManager.flush?.();
```
tt-a1i: [P2] Verify persistence before announcing that fullscreen was selected

The locked Pi SettingsManager records settings load/write errors internally; flush() waits for its queue but does not necessarily reject. With a malformed global settings.json, setTuiMode does not save, flush resolves, and this code still tells the user to restart to apply fullscreen. I reproduced this using the real SettingsManager: the invalid file remained unchanged, drainErrors() reported a global error, and the success/restart notification was emitted. Check the settings error/result and persisted readback before announcing success; report an actionable failure otherwise. The current mock cannot cover this behavior.

extensions/windows-tui-compatibility/index.ts
```diff
@@ -0,0 +1,133 @@
...
+    cleanup();
+    if (!shouldInstallWindowsTuiCompatibility(platform, ctx.mode)) return;
+
+    activeUi = ctx.ui;
+    ctx.ui.setWidget(
+      WIDGET_KEY,
+      (tui) => {
+        // The renderer can be replaced at runtime when the user switches TUI
+        // modes, so apply this when the factory receives the active renderer.
+        if (tui.mode === "regular") tui.setClearOnShrink(true);
```
tt-a1i: [P2] Respect an explicit terminal.clearOnShrink preference

For a user who explicitly selects regular mode and terminal.clearOnShrink:false, mounting this widget unconditionally flips the live renderer to true while the stored/native setting remains false. I reproduced that mismatch with SettingsManager.inMemory and the widget factory. This overrides an existing Pi rendering preference (including users reducing redraws on slow terminals), despite preserving their explicit regular mode. Apply the fallback only when the user has not explicitly configured clear-on-shrink, or require an explicit opt-in, and add a regression for the false setting.

tt-a1i (changes requested):
The Windows stale-autocomplete-row problem is worth fixing, but this implementation needs changes before merge. Please address the three inline findings: unsolicited global preference persistence, success notification after failed persistence, and overriding an explicit clear-on-shrink preference.

Validation: reviewed commit 441a4677c9ee2b38f75f9d37e72b3debe7b948b3; bun run check and the 4 focused tests passed in the earlier same-head validation; required CI is currently green. Re-ran controlled probes with the locked real Pi SettingsManager and a simulated win32 extension context to reproduce the findings. This is not a native Windows visual acceptance test. No merge performed.

zhouyuanxinand:
Addressed all three requested P2 findings in commit `4465e9d`:

- Removed the automatic global `tuiMode` write, persistence flush, and restart/success notification. The renderer workaround is session-local and leaves mode selection to Pi's native `/settings`, `settings.json`, or `--tui-mode` flow.
- Read global/project `terminal.clearOnShrink` before installing the workaround; an explicit value (including `false`) is preserved, and settings-read errors fail closed.
- Kept the workaround limited to the current interactive Windows regular renderer; fullscreen, RPC, print, JSON, and non-Windows sessions are unchanged.

All required PR CI jobs are green: Node 22, Node 24, and Background terminals (Windows).

extensions/windows-tui-compatibility/index.ts
zhouyuanxinand: Addressed in `4465e9d`. The extension no longer calls `setTuiMode`, `flush`, or any settings writer. It only applies the renderer workaround to the current interactive Windows session; persistent mode selection remains with Pi's native settings flow.

extensions/windows-tui-compatibility/index.ts
zhouyuanxinand: Addressed in `4465e9d`. The persistence path and success/restart notification were removed entirely, so a failed settings write can no longer be reported as successful. The extension now reads settings only and fails closed when the read reports errors.

extensions/windows-tui-compatibility/index.ts
zhouyuanxinand: Addressed in `4465e9d`. The fallback is enabled only when neither global nor project settings explicitly defines `terminal.clearOnShrink`; an explicit `false` (or `true`) is left untouched. The focused regression covers the explicit `false` case.

extensions/windows-tui-compatibility/index.ts
```diff
@@ -0,0 +1,135 @@
...
+      }
+    }
+
+    activeUi = ctx.ui;
+    ctx.ui.setWidget(
+      WIDGET_KEY,
+      (tui) => {
+        // The renderer can be replaced at runtime when the user switches TUI
+        // modes, so apply this when the factory receives the active renderer.
+        if (tui.mode === "regular" && enableClearOnShrink) {
```
tt-a1i: [P2] Apply the fallback when the native renderer actually changes

On Windows, start a session in fullscreen with clearOnShrink unset, then switch to regular through native /settings. This factory initially sees fullscreen and skips the override. In the locked Pi implementation, setExtensionWidget calls the factory only once and stores its returned component; switchTuiMode copies the previous clearOnShrink (default false) and remounts existing components without invoking the factory again. Since this component has empty render/invalidate methods, the regular renderer keeps false and the stale-row workaround never activates. Please handle the real renderer lifecycle and test this supported transition rather than simulating a second factory mount.

tt-a1i (changes requested):
复审当前 exact head。Windows 补全残影值得修复；本次改为会话内设置、保留显式偏好并在读取失败时不覆盖，已解决旧版相关问题。但真实 Pi 的 renderer 切换生命周期仍有一项 P2 功能缺口，需作者修改。

验证：精确 HEAD 的 5/5 单测通过；两次独立审查核对锁定 Pi 0.85.1 的 setExtensionWidget/switchTuiMode 实现。现有测试手动再次调用 widget factory，未覆盖原生模式切换。这是源码和接口语义验证，不是 Windows 实机视觉验收。未修改源码或合并。

extensions/windows-tui-compatibility/index.ts
zhouyuanxinand: Addressed in commit 79fdc90. The zero-height component now follows the real Pi renderer lifecycle: when a component created for fullscreen is remounted and invalidated after switchTuiMode swaps the active renderer, it re-registers through ctx.ui.setWidget. The factory therefore receives the current regular renderer and enables clearOnShrink there. The harness now mirrors the eager setExtensionWidget factory call plus component reuse and invalidation; it no longer manually mounts the factory a second time, and the transition regression also confirms an explicit clearOnShrink false remains untouched. Validation: bun run check passed; the focused suite passed 6/6 on native Windows and 6/6 in Ubuntu WSL2. The full local Windows suite exercised these 6/6 successfully but had unrelated parallel timeout and cancellation failures; representative timeout cases passed when rerun individually. All required PR checks are green: Node 22, Node 24, Node 26, Web E2E, and Background terminals (Windows).

tt-a1i (commented):
Reviewed exact head 79fdc900e7b8342d43c2b6770d2bf788bf3d23ac.

The latest patch does fix the earlier renderer-switch problem, preserves explicit startup configuration, fails closed on read errors, and cleans up per Session. However, #495 has now merged and closed #407. Merging this PR as-is would auto-load two extensions that both control clearOnShrink with different policies, while still not fixing the explicit terminal.clearOnShrink=false behavior in #495.

So the current PR is superseded and no longer has independent merge value. Please close it, or rework it as an explicit replacement that removes the #495 implementation/tests/docs and owns the behavior once. A smaller P2 remains in the current design: invalidate() is used as a lifecycle hook and can replace the widget during invalidation, relying on undocumented render ordering.

Exact-head tests, bun run check, and hosted CI are green; this is a competing-owner/value problem rather than a failing-test problem.

tt-a1i (commented):
审查提交：`79fdc900e7b8342d43c2b6770d2bf788bf3d23ac`。

当前主线已覆盖或存在同一职责的实现，建议核对后关闭本 PR。

### 验证范围

Complete diff inspected. Latest patch fixes earlier write-to-global and renderer-remount complaints; current main already auto-loads extensions/windows-terminal-compat/index.ts for the same clearOnShrink lifecycle.

### 限制与后续

Do not merge a second auto-loaded owner. Close or explicitly replace current implementation in a focused follow-up. No native Windows visual acceptance performed.

PR #414, opened 2026-09-06
fix(git-info): stop stale polling after session replacement

627150795:
## Problem

After rapid session replacement, overlapping `session_start` handlers could leave polling fibers using an invalidated Pi context. The stale context was then accessed every five seconds and logged a background defect indefinitely.

## Value

Stop stale polling without affecting the newest session's refresh loop, and explicitly interrupt polling during session shutdown.

## Approach

- Bind each polling fiber to the session generation and context that created it.
- Ignore superseded `session_start` handlers after their previous-fiber interruption completes.
- Interrupt the tracked polling fiber during shutdown.
- Add a regression test covering rapid session starts, stale-context access, and continued polling on the newest context.

## Validation

- `bun run check`
- `node --test --experimental-strip-types tests/extensions/git-info/index.test.ts tests/extensions/git-info/refresh-coordinator.test.ts`

The full suite still has one pre-existing Windows failure in `tests/extensions/git-info/process.test.ts`: a command exiting with code 7 is reported as -1.

## Impact

No public API changes. The change is limited to git-info background polling lifecycle and its regression coverage.

### DIFF
--- a/extensions/git-info/index.ts
+++ b/extensions/git-info/index.ts
@@ -158,18 +158,24 @@ export default function gitInfo(pi: ExtensionAPI) {
   const refresh = (ctx: ExtensionContext, forcePullRequest = false) =>
     refreshCoordinator.run(refreshEffect(ctx, forcePullRequest, generation));
 
-  const refreshIfIdle = (ctx: ExtensionContext) =>
-    refreshCoordinator.runIfIdle(refreshEffect(ctx, false, generation));
+  const refreshIfIdle = (
+    ctx: ExtensionContext,
+    refreshGeneration = generation,
+  ) =>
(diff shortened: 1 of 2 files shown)


tt-a1i (approved):
Reviewed the exact head and its unchanged patch integrated with main 561fd45. Two independent reviews found no actionable P2-or-higher issue. The session generation guard fixes stale polling without widening refresh concurrency; actual Pi shutdown sequencing was checked. Focused tests pass, and the combined #414/#436 integration passes bun run check and the full suite (1452 Node passes, 1 skip, 92 Vitest passes). Existing exact-head CI passes. Original author Adam Cheng is retained.

PR #418, opened 2026-09-06
feat(tui): add compact image paste placeholders

### LINKED ISSUE #527: feat(tui): compact image paste placeholders (display-only, relates to #413)
## Problem

Relates to #413.

Pasting clipboard images in the TUI editor inserts a long temporary file path that occupies most of the line width, makes the prompt unreadable when multiple images are attached, and requires dozens of backspace presses to remove. #413 identified this problem and proposed compact `[Image #N]` placeholders, which this issue carries forward with a narrower scope.

#413's acceptance criteria also include converting images to multimodal `ImageContent` at submission time and cleaning up temporary files. After investigating the codebase during the #418 implementation, those two items turn out to change submission semantics rather than improve display, and they conflict with Pi's existing design:

- **Submission semantics.** On the current `main` branch the only `input` handler that produces `ImageContent` is `extensions/ai-providers/cursor/input-images.ts`, gated on `ctx.model?.provider === "cursor"`. For every other provider, the model receives the raw file path and reads the image through the `read` tool on demand. Making all providers receive `ImageContent` unconditionally changes what the model sees relative to an unmodified Pi install.
- **Temporary file cleanup.** Pi leaves clipboard images in the temp directory so the `read` tool can open them in any later turn. The Cursor input-images implementation also does not delete files after conversion. Deleting on submission would break `read` access during compaction, rollback, and session reload.

Both items are worth pursuing but belong in a dedicated issue where provider-aware capability gating (`model.input.includes("image")`), the interaction with `downgradeUnsupportedImages`, and file lifecycle ownership can be discussed on their own terms.

## Proposed scope

This issue covers **TUI display and editing only**. The model receives the same bytes it would without this extension installed.

## Acceptance criteria

- [ ] Pasted clipboard images display as `[Image #N]` in the TUI editor.
- [ ] Backspace on any position within `[Image #N]` removes the entire token atomically.
- [ ] Multiple images in one input are numbered sequentially (`[Image #1]`, `[Image #2]`, …).
- [ ] Submission expands placeholders back to real file paths so the model and `read` tool receive unmodified content.
- [ ] The transcript renders submitted clipboard paths as compact placeholders (not raw temp paths).
- [ ] Transcript collapsing only rewrites files Pi wrote to the temp directory; paths in other directories, Markdown link/image targets, fenced code blocks, and inline code are left verbatim.
- [ ] Alt+Enter queued messages and Up-arrow history recall display compact placeholders and expand correctly on resubmission.
- [ ] Temporary files are not deleted by this extension; Pi's existing `read`-tool contract is preserved.
- [ ] Existing static checks and test suite remain passing.

## Non-goals (deferred)

- Converting pasted images to `ImageContent` for multimodal delivery.
- Temporary file lifecycle management (cleanup on placeholder deletion, session end, etc.).
- Extending compact placeholders to non-TUI surfaces (web, RPC).

These are tracked as potential follow-ups, not as blockers for this issue.

onychen:
## Problem

Closes #527. Relates to #413.

Pasting clipboard images in the TUI editor inserts a long temporary file path that occupies most of the line width, makes the prompt unreadable when multiple images are attached, and requires dozens of backspace presses to remove.

## Value

- Pasted images display as compact `[Image #N]` tokens while editing.
- Backspace anywhere inside a token removes it atomically.
- The transcript renders submitted messages with compact placeholders, not raw temp paths.
- Image numbering stays stable and intuitive after deletions.
- The model receives byte-identical content to an unmodified Pi install; the `read` tool can still open the file.

## Approach

The extension mirrors Pi's native long-paste marker pattern: the editor buffer shows a compact token, submission expands it back to the real content.

**Editing and submission**

- A new editor layer intercepts clipboard image pastes without modifying Pi's native editor.
- Only Pi's own clipboard images in the system temp directory are recognized; ordinary file paths are left alone.
- Pasting replaces the temp path with `[Image #N]` and tracks the mapping in a draft registry.
- `getExpandedText()` is overridden to expand placeholders back to real paths. Pi reads the submitted text through this seam before choosing a delivery path — `handleFollowUp()` calls it ahead of `prompt()`, `queueCompactionMessage()`, and `onSubmit` — so all three Alt+Enter branches expand correctly.
- Duplicate or user-edited placeholders lose their mapping to avoid one token claiming the wrong image.

**Transcript rendering**

- `registerMarkdownTransformer` collapses clipboard paths back to `[Image #N]` at render time, the same public seam Pi's built-in mermaid renderer uses.
- The transform is a stateless pure function scoped to user messages; replays, forks, and reloaded sessions render identically without any mapping surviving submission.
- Provenance-bounded: only files in known temp directories are collapsed. Paths in arbitrary project directories, Markdown link/image targets, fenced code blocks, and inline code are left verbatim.

**History and queued messages**

- Pi queues `getExpandedText()` results for Alt+Enter follow-ups and stores them in editor history. When these expanded paths reenter the editor through `setText` or `setTextInternal` → `onChange`, the extension detects them, rebuilds the placeholder mapping, and collapses the display. The next submission expands again correctly.

## Known limitation

Pending message display (`Follow-up: ...` in the queue area) is rendered by Pi's `updatePendingMessagesDisplay` using the queued text directly. The extension cannot intercept that rendering, so the queue area may briefly show the expanded path. This is a cosmetic limitation of the available extension seams and does not affect what the model receives.

## Validation

- 35 targeted tests pass, covering: placeholder editing, atomic deletion, submission expansion, numbering stability, consecutive submissions, all three Alt+Enter branches, `getExpandedText()` purity, transcript collapsing with provenance and Markdown awareness, dequeue restoration, history recall, adjacent path separation on both Windows and POSIX, **repeated identical paths through dequeue/history round-trips**, and **Windows usernames with spaces**.
- Type check, lint, and format checks pass (pre-existing failures in `web-store.ts`, `git-info`, and `pi-coding-agent-entry` are unrelated).
- Manual verification: editor shows `[Image #1][Image #2]`, transcript shows placeholders, model receives full paths, `read` tool opens the file, Up-arrow recall shows placeholders and resubmission expands correctly.

### Review round fixes (40af1e4)

Addresses both blocking findings from the 2026-09-13 review:

- **[P1] Repeated identical paths**: `adoptExpandedText` no longer coalesces repeated occurrences of the same clipboard path into a single placeholder mapping. Each `matchAll` occurrence gets its own placeholder ID and is spliced at its character index, so `expandWith`'s single-occurrence guard expands every copy on submission. Two new tests cover dequeue and history-recall round-trips with a repeated path.
- **[P2] Spaced Windows paths**: `CLIPBOARD_PATH_IN_TEXT` no longer excludes whitespace from directory segments (`[^\s\\/:"'<>|]+` → `[^\\/:"'<>|]+?`). Paths through usernames containing spaces (e.g. `C:\Users\Jane Doe\AppData\Local\Temp\...`) now match correctly in both transcript collapsing and editor adoption. Two new tests cover transcript collapse and adoption with a spaced Windows username.

## Impact

Display-only change to TUI interactive image pasting. Does not alter the content sent to the model, does not delete temporary files, and does not affect non-TUI surfaces.


extensions/image-paste/index.ts
```diff
@@ -0,0 +1,462 @@
...
+      removeTemporaryImage(attachment.path);
+    }
+  }
+
+  consumeSubmission(text: string) {
+    let submission =
+      [...this.pending.values()].find((candidate) => candidate.text === text) ??
+      [...this.pending.values()].find((candidate) =>
+        candidate.attachments.every((attachment) =>
+          text.includes(attachment.placeholder),
```
tt-a1i: [P1] Do not claim an older submission by placeholder text

A pending attachment can be sent with a later unrelated message. Reproduction: paste an image, submit it with Enter while compaction is active, and flush the native compaction queue with `willRetry: true`. Pi calls `session.steer(text)`/`followUp(text)` without emitting `input`, so the image remains in this store's pending map. Then submit ordinary text such as `Explain the literal token [Image #1]` without pasting any image. This fallback matches the old submission and attaches its image to that new message. I reproduced the native queue emitting only `[Image #1]`, followed by the unrelated text receiving the old image bytes. This goes beyond the documented compaction limitation: it changes which user message owns the image. Bind consumption to an actual submission identity and settle/discard attachment ownership across native queue paths; repeated placeholder text must not authorize claiming an older attachment. Please add a regression test covering compaction followed by an unrelated placeholder-containing message.

extensions/image-paste/index.ts
```diff
@@ -0,0 +1,462 @@
...
+      super.setText(text);
+    } finally {
+      this.settingText = false;
+    }
+    if (text.length > 0) {
+      this.attachments.reconcileDraft(text);
+    } else if (this.attachments.hasDraft) {
+      // Alt+Enter clears the editor immediately before submitting. Let the
+      // same-turn submit/input handler claim the draft before reconciling it.
+      setTimeout(() => this.attachments.reconcileDraft(this.getText()), 0);
```
tt-a1i: [P2] Preserve Alt+Enter attachments across asynchronous input handlers

The zero-delay timer can delete the image before the native input pipeline reaches this extension. Streaming Alt+Enter clears the editor and calls `session.prompt()` directly, bypassing `onSubmit`, so the attachment is still a draft. Pi's `ExtensionRunner.emitInput` awaits earlier handlers sequentially. With an earlier handler that awaits 10 ms, this timer reconciles the empty editor and deletes the image first; `transformImageAttachmentInput` then returns undefined, leaving `[Image #1]` in the message with no image and no warning. I reproduced this with the actual Pi 0.85.1 input dispatcher. Transfer ownership at the submission boundary, or use a lifecycle signal that distinguishes submission from an abandoned draft, instead of assuming input consumption finishes before the next timer tick. Please cover an asynchronous preceding input handler in the Alt+Enter regression test.

tt-a1i (changes requested):
Reviewed the current head 9c7f9ba6e5c9a32d859b69e093a41d72402c30d8. The compact placeholders and atomic deletion address a useful TUI problem, but attachment ownership still breaks across submission paths. Please address the two inline findings before merging.

Validation: the 11 image-paste/editor-layer tests pass locally, and all three required CI jobs are green. Additional minimal reproductions using Pi 0.85.1's actual InteractiveMode.flushCompactionQueue and ExtensionRunner.emitInput methods reproduce both findings. These are programmatic lifecycle reproductions, not visual TUI acceptance or live provider calls.

extensions/image-paste/index.ts
onychen: Fixed in 868b7cc. Alt+Enter now transfers draft ownership at the native app.message.followUp key-action boundary, before Pi clears the editor. It no longer relies on a zero-delay timer winning against ExtensionRunner. The regression waits for an asynchronous preceding input handler before transformation and verifies the image bytes remain available; ordinary draft clearing still removes the temporary file.

extensions/image-paste/index.ts
onychen: Fixed in 868b7cc. Pending attachments are now consumed only from the ordered submission boundary with an exact original-text guard; placeholder containment can no longer select an older submission. The session_compact(willRetry) path explicitly discards pending ownership and removes its temporary files before Pi flushes through steer()/followUp(). Added regressions for both an identical later [Image #1] after compaction retry and a different message containing the same placeholder.

tt-a1i (changes requested):
Reviewed at head bc2739f9228242d3c0c855823a3c9d6316893d14.

## Standards

[P1] Image ownership is inferred only from the OS temp directory, a pi-clipboard filename pattern, and statSync(). statSync follows symlinks, and the later read/delete has a check-to-use race. The extension can therefore claim and remove a matching file without runtime-proven ownership. Please use a Pi-provided attachment handle, or copy into an OpenPI-owned directory and validate identity with no-follow, bounded I/O.

[P1] The 481-line editor layer depends on several private lifecycle details at once: onSubmit wrapping, raw Alt+Enter handling, setText cleanup timing, input ordering, and compaction retry behavior. Pi 0.85.1 still exposes no stable attachment/submission seam, so upgrades can silently break ownership. This should be reduced around an explicit Pi-native attachment/submission boundary rather than duplicating InteractiveMode lifecycle assumptions.

[P2] The input handler performs unbounded synchronous readFileSync plus base64 conversion. A large clipboard file can block the TUI and amplify memory use. Add a size limit and bounded asynchronous reading after identity validation.

## Spec

[P1] Successful compaction with multiple queued submissions is still unsafe. Every submission enters pending, but after normal compaction only the first queued message goes through prompt/input; later messages use steer/followUp and bypass the input transform. Their images are neither sent nor promptly cleaned, and a later identical text can consume an old FIFO submission. The existing test covers only one willRetry=true submission, not successful compaction with multiple queued submissions or repeated identical text.

The Alt+Enter cleanup race from the previous review is fixed, but attachment ownership across compaction remains incomplete. Please add the missing lifecycle evidence and resolve the current main conflict before requesting re-review.

onychen:
Thanks for the review. I've reworked the approach rather than patching the individual findings, because the structural concern turned out to be the root of the others.

**What changed**

The previous version converted placeholders into base64 `ImageContent` inside the `input` event and deleted the temp file after reading. That meant owning an attachment lifecycle that Pi does not expose a stable seam for, which is what forced the dependency on private submission internals.

This version follows Pi's own long-paste marker pattern instead: the buffer shows a compact marker, submission expands it back to the real content. Concretely, the extension no longer reads files, no longer produces base64, and no longer deletes anything. The model receives the real clipboard path, byte-identical to unmodified Pi, and the `read` tool can still open it.

`Submission`, the `pending` map, and the `input` and `session_compact` handlers are all gone.

**How that maps to your findings**

- Spec P1 (images orphaned when a successful compaction flush dispatches through `steer`/`followUp`): no attachment state crosses the submission boundary anymore, so there is nothing to orphan or misclaim. Expansion now happens in `getExpandedText()`, which `handleFollowUp()` calls before it picks between `prompt()`, `queueCompactionMessage()` and `onSubmit` — so all three Alt+Enter branches expand correctly. Regression tests cover each branch.
- Standards P2 (unbounded synchronous read): no file is read at submission time.
- Standards P1 (ownership determination): the extension no longer takes ownership of the temp file, so the symlink/TOCTOU surface is gone.
- Standards P1 (editor layer depends on too much private lifecycle): the assumptions collapse to a single documented seam, `getExpandedText()`. I verified against 0.85.1 that Pi routes every path needing the submitted text through it.

The main conflict is also resolved.

**One thing worth flagging**

Expanding at submission solved the model side but relocated the problem the issue actually reported: the long temp path moved out of the editor and into the transcript. I've added a rendering-only counterpart via `registerMarkdownTransformer` — the same public seam the built-in mermaid renderer uses — which collapses clipboard paths back to `[Image #N]` for user messages. It's a stateless pure function, so replays, forks and reloaded sessions all render consistently without any mapping surviving submission.

While testing that I found and fixed a real bug: pasting images back to back produces adjacent paths with no separator, and a greedy directory match absorbed the following path's `C:` or `/tmp`, rendering several images as one placeholder. Covered by tests for both Windows and POSIX shapes.

**Known limitation**

Transcript numbering follows the order paths appear in the message, so it can differ from the editor if images were deleted mid-draft. Making it strictly identical would require carrying a mapping through message metadata, which is outside the seams available to an extension. Since transcript numbering only distinguishes images within a single message, I've documented it rather than worked around it.

Longer term, folding image paths into Pi's native paste marker registry seems like the better home for this. Happy to open an upstream issue if you agree.

24 targeted tests pass; type check, lint and format are clean.

tt-a1i (changes requested):
Reviewed exact head 52abcea852e8d927644510e9d3e23871ed7e41bf.

The redesign is substantially simpler and fixes the earlier ownership/TOCTOU, unbounded read, private submission-state, and compaction-loss problems. Two P1 acceptance gaps remain:

- [P1] #413 requires submitting the attachment as multimodal ImageContent and cleaning up temporary resources. This implementation expands the token to ordinary filesystem-path text, and cleanup() explicitly preserves the file. The model receives no image payload unless it independently reads the path. Either the accepted spec must change explicitly or this needs a Pi-native attachment seam and lifecycle.
- [P1] Pi queues the result of getExpandedText() for Alt+Enter follow-ups. The pending display therefore shows the full temp path, and Alt+Up restores that path through setText() without rebuilding the attachment mapping. The compact multi-input UX still breaks on the native queued-message lifecycle; current tests do not cover pending display, dequeue, or history recall.

[P2] collapseClipboardPaths() also rewrites any matching basename in any directory, without provenance or Markdown awareness, so an unrelated path can be hidden and a Markdown image target can be corrupted.

Focused tests and hosted CI are green, but the two P1s block merge.

onychen:
Thanks for the thorough review. Addressing each point against the new head:

**[P1] ImageContent and file cleanup — scope change**

After investigating `main`, the only `input` handler that produces `ImageContent` is `cursor/input-images.ts`, gated on `ctx.model?.provider === "cursor"`. For every other provider the model receives the raw path and reads the image through the `read` tool. Neither Pi nor the Cursor implementation deletes the temp file.

Making all providers receive `ImageContent` unconditionally changes submission semantics and conflicts with Pi's `read`-tool contract. These are valid goals but belong in a dedicated issue where provider-aware gating (`model.input.includes("image")`), the interaction with `downgradeUnsupportedImages`, and file lifecycle ownership can be discussed properly.

I've opened #527 with acceptance criteria scoped to TUI display and editing only — the model receives byte-identical content to an unmodified Pi install. The PR description now targets #527 and the rationale is documented there. #413 has been commented with the split explanation.

**[P1] Queued messages, dequeue, and history recall — fixed**

Three changes in this push:

1. `setText()` now calls `adoptExpandedText()`: when Pi restores queued messages through `restoreQueuedMessagesToEditor` → `setText()`, clipboard paths in the text are detected, the placeholder mapping is rebuilt, and the buffer shows compact tokens. The next submission expands correctly.

2. `onChange` also calls `adoptExpandedText()` when the draft is empty: Pi's history recall uses `setTextInternal()` which bypasses `setText` and only fires `onChange`. The `!this.attachments.hasDraft` guard ensures this only runs on externally injected text (recall, dequeue), not during normal editing.

3. `addToHistory` is *not* overridden — history stores the real path (matching Pi's native behavior). When Up-arrow recall writes it back through `setTextInternal` → `onChange`, point 2 above adopts and collapses it.

Pending display (`Follow-up: ...` in the queue area) is rendered by Pi's `updatePendingMessagesDisplay` using the queued text directly — there is no extension seam to intercept it. This is documented as a known limitation in the PR description; it does not affect what the model receives.

New tests cover: dequeue restoring placeholders and expanding again, multi-image dequeue renumbering, and history recall via the `setTextInternal` → `onChange` path.

**[P2] collapseClipboardPaths provenance and Markdown awareness — fixed**

Three guards added:

- **Provenance**: only files in known temp directories are collapsed (`tmpdir()` plus portable POSIX/Windows temp path patterns). A project file that shares the clipboard basename keeps its real path.
- **Code spans**: fenced blocks (`` ``` ``/`~~~`) and inline code are detected and skipped, so paths inside code remain copy-pasteable.
- **Link/image targets**: `](path)` positions are detected and skipped, so Markdown image references are not corrupted.

New tests cover: non-temp directory paths preserved, Markdown image targets preserved, fenced and inline code preserved, and prose still collapses when the same message contains a code block.

extensions/image-paste/index.ts
```diff
@@ -0,0 +1,525 @@
...
+  adoptExpandedText(text: string) {
+    const paths = [...text.matchAll(CLIPBOARD_PATH_IN_TEXT)]
+      .map((match) => match[0])
+      .filter((path) => isPiClipboardPath(path));
+    if (paths.length === 0) return undefined;
+
+    this.clearDraft();
+    let collapsed = text;
+    for (const path of paths) {
+      if ([...this.draft.values()].some((entry) => entry.path === path)) {
```
tt-a1i: [P1] Do not coalesce repeated path occurrences into one placeholder mapping. replaceAll() turns every copy of the same queued/history path into the same token, but expandWith() deliberately expands only tokens occurring exactly once. Dequeuing two messages that reference the same clipboard path therefore displays [Image #1] twice and resubmits those literal tokens instead of the original paths. Give each occurrence its own mapping/token or preserve a reversible occurrence-level mapping, and add a dequeue/history round-trip test with a repeated path.

extensions/image-paste/index.ts
```diff
@@ -0,0 +1,525 @@
...
+ *
+ * Pasting two images produces adjacent paths with no separator between them,
+ * so the directory portion is matched as discrete segments that exclude both
+ * path separators and `:`, and it is lazy rather than greedy. A greedy middle
+ * absorbs the next path's `C:` or `/tmp` and renders the pair as one
+ * placeholder; laziness ends each match at the first filename that completes
+ * it, which is exactly the boundary between two pasted images.
+ */
+const CLIPBOARD_PATH_IN_TEXT =
+  /(?:[A-Za-z]:[\\/]|[\\/])(?:[^\s\\/:"'<>|]+[\\/])*?pi-clipboard-[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}\.(?:gif|jpe?g|png|webp)/gi;
```
tt-a1i: [P2] This scanner excludes whitespace from every directory segment. On Windows, a normal temp directory such as C:\Users\Jane Doe\AppData\Local\Temp therefore never matches, so the transcript and dequeue/history recall expose raw paths instead of placeholders. Please parse against the known temp directory (or otherwise support spaced segments) and cover a Windows username or TMPDIR containing spaces.

tt-a1i (changes requested):
Reviewed exact head b1ab179ef9dfcfb9dcb6c4a0779eedb9ca975536.

The display-only redesign is a much better fit for Pi's current extension seams and resolves the previous attachment-ownership, TOCTOU, synchronous-read, and compaction-loss findings. Two acceptance gaps remain:

- Repeated instances of the same clipboard path do not survive the dequeue/history round trip. Adoption coalesces them into one placeholder identity, while submission refuses to expand a placeholder that occurs more than once, so literal [Image #1] tokens reach the model instead of the paths.
- The transcript/adoption scanner rejects whitespace in directory segments, so ordinary Windows temp paths for usernames containing spaces are not collapsed or re-adopted.

Validation: bun run check passed; the full branch suite passed with 1496 Node tests and 1 skip plus 130 Vitest tests. A no-conflict merge onto current main also passed bun run check and all 32 focused image-paste/editor-layer tests. All six hosted checks are green. Both findings above have minimal programmatic reproductions; this was not a visual TUI acceptance run.

The PR is still draft and behind main. Please fix these two cases and update from main before requesting re-review.

JS-banana:
image

Verified on a real TUI as the #413 reporter: pi 0.85.1 on macOS, loading this branch via `pi -ne -e extensions/image-paste/index.ts`, inside a GUI terminal host (Orca) — pi's Ctrl+V paste reads the system clipboard natively, so the path is host-independent.

Checked against the #527 criteria: paste shows `[Image #1]`; backspace anywhere in the token removes it whole; multiple pastes number sequentially; submission sends real paths and the model read the image back correctly; transcript renders placeholders; Alt+Enter dequeue and ↑ history recall both restore compact placeholders and expand on resubmit.

Also +1 to the upstream direction you mentioned earlier (folding image paths into Pi's native paste marker registry). For reference, oh-my-pi (can1357/oh-my-pi) validates that form: its editor exposes a generic atom table (`insertAtom` + `atomicTokenPattern`) which gives atomic deletion, undo, and draft persistence natively, and the wire format persists `[Image #N, WxH]` so editor and transcript numbering share one source. An equivalent upstream seam would dissolve both known limitations documented here.

tt-a1i (approved):
Re-reviewed exact head 2a7f25c9b1cac49f1418453df9a899c724bb8af1.

The two remaining acceptance failures are resolved. Repeated occurrences of the same clipboard path now receive independent placeholder identities during dequeue/history adoption and expand back to every original occurrence. Windows temp paths with spaces in the username are recognized across transcript rendering and editor adoption.

Validation: 36 focused image-paste/editor-layer tests passed; an additional direct reproduction covered Windows spaced-path adoption combined with repeated occurrences; merging the PR head with current main was conflict-free and bun run check passed. Hosted CI is green. The #413 reporter also recorded real Pi 0.85.1 TUI acceptance for paste, atomic deletion, multiple images, model/read delivery, transcript rendering, dequeue, and history recall.

No blocking findings.

Issue #424, opened 2026-09-06
fix: make child tool, workspace and provider execution reliable


tt-a1i:
## Problem and requested behavior

Ordinary and Workflow children should be able to use parent-authorized tools, including shell/network tools, and operate in an explicitly selected working directory. A recent eight-way PR audit failed across three attempts: Cursor chat-only rejected tool calls; Codex children were stopped by OpenPI's extra 45-second progress watchdog; Seal returned account rate limits while the task/tool/cwd mismatch persisted. Private sessions and credentials are not published.

The user authorizes investigation and implementation in one PR, for review before merge.

## Scope / acceptance

- Built-in role defaults support parent-authorized child-safe tools. Custom allowlists and parent-only exclusions remain enforceable; Plan Mode cannot obtain write/shell authority through a child.
- Ordinary and Workflow paths have tested explicit cross-directory operation, with target-project trust and correct worktree/replay/operator/cleanup identity.
- Cursor maps supported advertised tools to Pi's native tool lifecycle and replays results; it must not execute Cursor-native tools outside Pi. Local protocol fixtures plus bounded real acceptance distinguish proven support from uncertainty.
- Remove false 45-second child failures without weakening Pi transport error handling or bounded caller cancellation/disposal.
- Inspect related observed hot paths; add focused regressions and executable evidence rather than unrelated refactoring.
- `bun run check`, `bun run test`, independent reviews and required CI pass.
- Submit one cohesive PR; do not merge or publish a release.

## Related work

- #420 / #422 fixed repeated dashboard history hydration; configured local runtime was still at `c8f2c13` during diagnosis.
- #419 / #423 track configured shell propagation; inspect overlap instead of duplicating that independent work.
- #185 / #327 track viewport transcript performance; inspect relevance before adding changes.

Repository investigation and design records will be linked from the delivery PR and this Issue.

seekskyworld:
I am taking a focused reliability slice from this issue: align the Workflow child model-progress watchdog with Pi's effective `SettingsManager.getHttpIdleTimeoutMs()` value. The current code reads only raw optional settings, so an unset Pi setting (whose effective default is 300 seconds) silently falls back to OpenPI's extra 45-second timeout and can abort a healthy slow provider turn.

[...]

PR #427, opened 2026-09-06
fix(workflows): follow Pi effective child idle timeout

### LINKED ISSUE #425: fix(workflows): honor Pi effective HTTP idle timeout for child progress watchdog
## Problem

Workflow child sessions currently derive `MODEL_PROGRESS_TIMEOUT_MS` from the raw optional `httpIdleTimeoutMs` fields in project/global settings. When the user has no explicit override, those fields are absent even though Pi's effective setting is its 300-second default. OpenPI therefore applies an undocumented 45-second provider-progress watchdog to every child, which can abort slow but healthy provider turns before Pi's own transport timeout.

This is a focused reliability slice of #424. It does not change tool execution, provider protocol support, or cancellation semantics.

## Evidence

- Pi 0.85.1 exposes `SettingsManager.getHttpIdleTimeoutMs()`, returning the effective default of 300000 ms.
- `extensions/workflows/runner.ts` currently reads `getProjectSettings()`/`getGlobalSettings()` directly and falls back to 45000 ms when both raw fields are absent.
- The existing watchdog remains useful as a bounded child guard, so an explicit Pi timeout of `0` will retain the OpenPI minimum rather than creating an unbounded child.

## Acceptance

- Resolve the watchdog default from Pi's effective HTTP idle timeout API.
- Preserve the 45-second minimum for effective values below that floor, including explicit `0` (disabled transport timeout).
- Preserve test-only overrides and existing abort/cleanup behavior.
- Add deterministic regression coverage for the effective default, wider values, floor values, and explicit zero.
- Update the user-facing timeout contract and link this issue to #424.

seekskyworld:
## Problem

Closes #425. Workflow child sessions derive their model-progress watchdog from raw optional `httpIdleTimeoutMs` fields. When no override is configured, those fields are absent even though Pi's effective default is 300 seconds, so OpenPI can abort a healthy slow provider turn after 45 seconds. This is an independently reviewable reliability slice of #424; it does not claim to solve Cursor tool support, cross-directory authorization, or allowlist behavior.

## Value

Child provider turns now use the same effective idle-timeout contract as Pi's native transport. Unconfigured sessions no longer inherit an undocumented 45-second watchdog, while the bounded OpenPI guard remains in place for low values and an explicit transport timeout of `0`.

## Approach

- Resolve the watchdog timeout through `SettingsManager.getHttpIdleTimeoutMs()` so Pi's merged settings and default are honored.
- Preserve the 45-second minimum, test-only overrides, and existing abort, progress-refresh, completion, and cleanup behavior.
- Add deterministic coverage for the 300-second effective default, wider explicit values, overrides, the floor, and explicit zero.
- Update the README contract and add the validated research record [`docs/research/ISSUE_425_WORKFLOW_WATCHDOG_2026-09-07.md`](https://github.com/openpi-dev/openpi/blob/6f3a26b/docs/research/ISSUE_425_WORKFLOW_WATCHDOG_2026-09-07.md).

## Validation

- `npx --yes bun@1.3.14 run test` — Node: 1430 passed, 1 skipped, 0 failed; Vitest: 30 passed.
- `npx --yes bun@1.3.14 run typecheck` — passed.
- `npx --yes bun@1.3.14 run lint` — passed.
- Targeted Biome format and `git diff --check` — passed.
- `npx --yes bun@1.3.14 run check` — the repository check reaches the formatter and reports only the pre-existing untracked `.golutra/local.json` and `.golutra/workspace.json` differences; config contract, discipline, Web syntax, lint, and typecheck pass. Those collaboration files were preserved.

## Impact

- User-visible behavior: slow healthy provider turns receive the Pi-configured idle window instead of an unintended 45-second cutoff.
- Runtime/lifecycle: the child progress watchdog remains bounded and keeps its existing abort and cleanup semantics.
- Persisted config/data: none; the resolver reads Pi's existing settings and does not write them.
- Compatibility/risk: explicit Pi values at or above 45 seconds are honored; values below 45 seconds and `0` retain the OpenPI floor. Cursor, workspace, and allowlist behavior are unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,7 +275,7 @@ return agent("Synthesize the verified findings", {
 | `pipeline()` | 每个 item 完成上阶段后立即进入下一阶段；多阶段 fan-out 的默认选择          |
 | `parallel()` | 并发 barrier；只在下一阶段确实需要全部结果时使用                           |
 
-Workflow 默认并发 8 个 Agent，单次最多 128 次调用；可配置到 64 和 1024。前台运行可实时查看，后台运行完成后自动回传；`/workflows` 展示阶段、Agent、Transcript、Graph、用量与产物。每个 Child Provider turn 必须在 45 秒内产生模型可见的 thinking、text、tool call 或完成事件，并在持续输出时按进展续期；空 stream start 与 transport heartbeat 不算进展。用户显式配置了更宽的 Pi `httpIdleTimeoutMs` 时沿用该上限。超时会 abort 当前 Child、保留已有 Transcript/usage/evidence，并让 sibling 与后续阶段继续结算。
+Workflow 默认并发 8 个 Agent，单次最多 128 次调用；可配置到 64 和 1024。前台运行可实时查看，后台运行完成后自动回传；`/workflows` 展示阶段、Agent、Transcript、Graph、用量与产物。每个 Child Provider turn 的进度 watchdog 跟随 Pi 的有效 `httpIdleTimeoutMs`（当前默认 300 秒），并保留 45 秒最低边界；它必须产生模型可见的 thinking、text、tool call 或完成事件，并在持续输出时按进展续期。空 stream start 与 transport heartbeat 不算进展。即使 Pi 的传输超时显式设为 `0`，子会话仍保留这个有界最低值。超时会 abort 当前 Child、保留已有 Transcript/usage/evidence，并让 sibling 与后续阶段继续结算。
 
 ---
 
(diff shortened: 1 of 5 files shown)


seekskyworld:
No maintainer review comments were present on this PR. After it was opened, upstream merged #426 (https://github.com/openpi-dev/openpi/pull/426, commit `eaf470b`), which removes the extra child output watchdog and delegates liveness to Pi's native transport timeout/retry lifecycle. That broader fix supersedes this focused `getHttpIdleTimeoutMs()` implementation; rebasing it onto current `main` (`b0fc558`) would reintroduce the removed 45-second watchdog. Closing this PR as superseded. The investigation and validation remain traceable in commits `26305d4` and `6f3a26b`.

PR #495, opened 2026-09-08
Codex/fix tui stale autocomplete 407

### LINKED ISSUE #407: fix(tui): prevent stale autocomplete rows on Windows terminals

REI-330:
## Problem

Windows regular TUI 在 slash 命令自动补全列表缩小时可能残留旧行，影响界面可读性。

## Value

避免 Windows 用户看到过期的自动补全内容，恢复稳定的终端重绘行为。

## Approach

- 新增 Windows TUI 兼容层，仅对 regular TUI 启用 Pi 的 `clearOnShrink`。
- fullscreen TUI 和非 Windows 平台保持原有行为。
- 提供 `OPENPI_WINDOWS_TUI_COMPAT=0` 环境变量作为退出开关。
- 添加针对平台、渲染模式和退出开关的回归测试。

## Validation

- `bun test tests/extensions/windows-terminal-compat/index.test.ts`：3/3 通过
- Biome 检查通过
- TypeScript 检查通过
- 完整检查受本地 Windows 环境既有 CRLF、权限和进程基线测试影响

## Impact

仅改变 Windows regular TUI 的终端重绘策略，不改变命令注册、编辑器行为、fullscreen TUI 或其他平台行为。

Fixes #407

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -64,6 +64,12 @@ Capability discovery defaults to `explicit`, preserving the zero-resident OpenPI
 
 OpenPI Web theme defaults to `system`; `light` and `dark` are explicit canonical setup choices, and the browser consumes them from each authoritative snapshot without writing a competing local preference.
 
+On Windows, OpenPI enables Pi's `clearOnShrink` compatibility behavior for
+the regular TUI so shrinking slash-command autocomplete lists do not leave
+stale rows on screen. Fullscreen TUI keeps its configured behavior. Set
+`OPENPI_WINDOWS_TUI_COMPAT=0` before starting Pi to opt out of this
+Windows-only adjustment.
+
 Legacy `footerItems` is accepted and migrated at the input boundary, but new setup writes persist only canonical `footerLines`. Configurations written by this version are not guaranteed to retain their Footer layout when read by an older OpenPI version.
 
 <!-- config-contract: capabilities.discovery suggestions.enabled suggestions.model workflows.concurrency workflows.maxAgentCalls ui.webTheme ui.showHeader ui.customFooter ui.footerStyle ui.footerLines ui.subagentResultDisplay ui.bashToolDisplay ui.fileMutationDisplay postEdit.command subagents.roleModels -->
--- a/SETUP.md
+++ b/SETUP.md
@@ -66,9 +66,8 @@ OpenPI Web theme defaults to `system`; `light` and `dark` are explicit canonical
 
 On Windows, OpenPI enables Pi's `clearOnShrink` compatibility behavior for
 the regular TUI so shrinking slash-command autocomplete lists do not leave
-stale rows on screen. Fullscreen TUI keeps its configured behavior. Set
-`OPENPI_WINDOWS_TUI_COMPAT=0` before starting Pi to opt out of this
-Windows-only adjustment.
+stale rows on screen. Fullscreen TUI keeps its configured behavior. The
+native `PI_CLEAR_ON_SHRINK=0` setting remains an explicit opt-out.
 
 Legacy `footerItems` is accepted and migrated at the input boundary, but new setup writes persist only canonical `footerLines`. Configurations written by this version are not guaranteed to retain their Footer layout when read by an older OpenPI version.
 
--- a/extensions/windows-terminal-compat/index.ts
+++ b/extensions/windows-terminal-compat/index.ts
@@ -13,7 +13,6 @@ type CompatibleTui = {
   mode: "regular" | "fullscreen";
   getClearOnShrink(): boolean;
   setClearOnShrink(enabled: boolean): void;
-  addInputListener(listener: (data: string) => unknown): () => void;
 };
 
 export function shouldClearShrunkRows(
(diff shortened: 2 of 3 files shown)


tt-a1i (changes requested):
Reviewed at head 6f102805b7ebaf234657a8a7c8958a34a2137d99.

## Standards

[P1] The new user-facing OPENPI_WINDOWS_TUI_COMPAT opt-out bypasses the package configuration contract. Package-owned feature toggles must use the canonical /openpi-setup path and update extensions/setup/, extensions/shared/setup-config.ts, the no-argument status output, SETUP.md, and README. This PR only adds the environment variable and SETUP.md documentation. Please either integrate the complete configuration path or remove the public toggle.

[P3] The same environment-variable check is repeated in install() and the editor wrapper. Once install() has passed, the second expression is always true.

## Spec

[P2] applyWindowsTerminalCompatibility() unconditionally sets clearOnShrink=true in Windows regular mode, overriding an explicit terminal.clearOnShrink=false preference and leaving persisted configuration inconsistent with runtime behavior.

[P2] The compatibility layer runs only when the editor factory is created. Pi 0.85.1 does not rerun that factory when switching from fullscreen to regular mode, so that supported lifecycle can still retain clearOnShrink=false and reproduce the stale-row bug. The current tests cover only the pure helper, not an actual mode transition.

Please preserve explicit user configuration and add a lifecycle regression for fullscreen to regular before merging.

tt-a1i (changes requested):
Re-reviewed at head bd4b76e52011f4966257f14642192b14dbc5a6af.

The custom OpenPI environment toggle and its configuration-contract issue are fixed. Two runtime blockers remain.

## Standards

[P1] Please register the listener through the Pi extension lifecycle instead of calling tui.addInputListener() directly. The tui value passed to the editor factory is a stable proxy, but addInputListener() returns a subscription bound to the current underlying renderer. Pi replaces that renderer when switching TUI modes and only rebinds subscriptions registered through ctx.ui.onTerminalInput(). As written, the listener remains attached to the old fullscreen renderer and the new regular renderer has no listener. It also retains the old renderer until shutdown.

A reproduction using the Pi 0.85.1 stable-TUI proxy produced:

    {"oldListeners":1,"newListeners":0,"clearOnShrink":false}

The new test changes mode on one mock object, so it does not model renderer replacement and gives a false positive. Please test with two renderer instances or the real Pi lifecycle.

[P3] Listener registration should be gated to Windows with the opt-out disabled. The current code installs a per-keystroke listener on Linux, macOS, fullscreen-only, and opted-out sessions even though it can never change behavior there. The PI_CLEAR_ON_SHRINK expression is also duplicated between initial application and listener installation.

## Spec

[P2] An explicit terminal.clearOnShrink=false setting is still overwritten on the next input. Pi SettingsManager preserves the provenance only before it reaches TUI; TUI.getClearOnShrink() returns false for both the default and an explicit false value. applyWindowsTerminalCompatibility() therefore cannot distinguish them and sets the runtime value back to true, leaving persisted configuration false while runtime behavior is true.

The test named preserves an already enabled or explicitly disabled renderer setting mocks getClearOnShrink() as true, so it only covers the already-enabled case. Please add the explicit-false case and preserve the native setting, including changes made through /settings.

tt-a1i (approved):
Re-reviewed at exact head 2a56557a1fa1cdb442868aa33ff99bb9f3bd4896. The previous P1 renderer-lifecycle blocker is fixed: the compatibility listener now uses `ctx.ui.onTerminalInput()`, so Pi rebinds it after renderer replacement. Focused tests (5/5), `bun run check`, merge-tree validation against current main, and hosted CI are green. No P0/P1 remains, and the fix is isolated to Windows TUI rendering rather than model context or Session semantics. Non-blocking P2: persisted `terminal.clearOnShrink=false` is still indistinguishable from the default false and can be forced back to true unless `PI_CLEAR_ON_SHRINK=0` is set; the test named for explicit native false currently models the environment opt-out instead of the persisted setting.

Issue #512, opened 2026-09-08
bug(subagents): --print/--json parent can settle while required children are still running [under investigation]


tt-a1i:
## Status

Wiring landed on research branch `research/openpi-cursor-reliability-20260908` @ `7950570` and [PR #515](https://github.com/openpi-dev/openpi/pull/515). **PR is not merged.** Isolated evidence only. Not a formal Benchmark.

## Impact

In `pi --print` / `--mode json`, a parent that calls `subagent_spawn` and then ends the turn emits `agent_settled` and the process exits. Automatic result delivery only helps an interactive TUI. Required child results are lost.

## Evidence — same E task+acceptor, `cursor-grok-4.6-high-fast`

Receipts still say `source=a3edee28` because the harness hardcoded that SHA. Split is by spawn-result text + attempt window. Index: `/tmp/openpi-reliability-20260908/results/e-high-fast-512-version-split.md`.

| Window | Product behavior | n | Correct |
| --- | --- | --- | --- |
| E-01..10 | stock v0.8.1 spawn (“end your turn” / auto re-invoke). E-06..10 are still this window, not after-fix. | 10 | 7 |
| E-11..20 | print-host spawn from `7950570` (“MUST call subagent_wait”). All 10 stdout traces show that text. | 10 | 10 |

E-01 spawn result (trimmed): “end your turn; its result is delivered automatically and you are automatically re-invoked”. E-11 spawn result: “This host cannot re-invoke you after you end the turn. You MUST call subagent_wait…”

Isolated `bun run check` exit 0 (`results/isolated-bun-check.log`) and `bun run test` exit 0, 1470 pass (`results/isolated-bun-test.log`).

## Fix (on the research branch only)

- `buildSubagentSpawnResult({ canDeliverLater: false })` requires wait and does not say “end your turn”.
- On `agent_settled`, if the host cannot deliver later and direct children are still running, send up to two `followUp` + `triggerTurn` barriers.

Do not disable tools, loosen permissions, or lengthen timeouts.

## Applicability

OpenPI v0.8.1 subagent spawn/delivery. Not Cursor native-tool rejection (PR #484). Related hung-after-success: #516.

tt-a1i:
Isolated worktree now has a failing-repro-first helper plus wiring (not merged):

- `extensions/subagents/src/print-host.ts` — TUI-only `canDeliverLater`, pending follow-up text
[...]

PR #515, opened 2026-09-08
fix(subagents): keep print-host parents from settling without wait

tt-a1i:
## Summary
- `--print` / `--mode json` cannot re-invoke the parent after `agent_settled`. Stock spawn text still tells the model it may end the turn; required child results are lost (E high-fast 3/10 print-settle failures).
- Non-TUI hosts now get spawn text that requires `subagent_wait`, plus a bounded `followUp` + `triggerTurn` barrier if the parent settles while direct children are running. Interactive TUI guidance is unchanged.
- Same E task+acceptor on `cursor-grok-4.6-high-fast`: **7/10 before** (v0.8.1 stock) → **10/10 after**. Isolated `bun run check` and `bun run test` pass. Closes nothing; tracks #512. Do not merge from this reliability session unless a maintainer asks.

## Test plan
- [x] `tests/extensions/subagents/print-host.test.ts` and prompt contract
- [x] `bun run check` && `bun run test` in isolated v0.8.1 worktree
- [x] Real-model E high-fast n=10 after on the same acceptor
- [ ] Reviewer: confirm TUI fire-and-forget still says “end your turn”
- [ ] Reviewer: confirm print/json spawn result does not

### DIFF
--- a/docs/research/CURSOR_RELIABILITY_BASELINE_2026-09-08.md
+++ b/docs/research/CURSOR_RELIABILITY_BASELINE_2026-09-08.md
@@ -1,0 +1,122 @@
+---
+status: draft
+created: 2026-09-08
+last-verified: 2026-09-08
+applies-to: isolated v0.8.1 worktree a3edee28 plus local OPENPI_CURSOR_WIRE_LOG instrumentation; not the user's daily runtime
+related-issues: "#234"
+related-prs: "#484, #485"
+supersedes: none
+---
+
+# OpenPI Cursor reliability experiment baseline
+
+Exploratory research setup and first live samples. Not a formal Benchmark.
+
+## Question
+
+Can parent-agent and subagent tasks on Cursor Grok configs be systematically accepted against machine-checkable acceptors on released OpenPI v0.8.1, and which failure classes are OpenPI-fixable?
+
+## Observed facts
+
+### User checkout (untouched)
+
+Commands run in `/Users/tushaokun/work/my-pi-setup` without writing to it:
+
+- Branch: `codex/knowledge-evidence-contract`
+- HEAD: `a9b40f0044ee59c360a6077c1c0bdbdbd30da10b`
+- Dirty: yes (docs/research, architecture visuals, `.hive/`, repro scripts). Left intact.
+- Local `main`: `2a69d3f32994da4123f1312b7fa84ef3d6119be1` (stale vs origin)
+- `origin/main`: `0d17f4577fe31315fe6c95370d251bdb4e2413cf` (contains v0.8.1)
+
+### Release identity
+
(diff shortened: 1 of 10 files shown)


tt-a1i (commented):
Self-review at exact head a10103da192b64d9d00e297a361ba9d396e4ec1f. I cannot formally request changes on a PR authored by this account, but the following are merge-blocking.

## Standards

[P1] Print-host completion is still prompt-enforced rather than runtime-enforced. The `agent_settled` handler injects at most two “MUST call subagent_wait” follow-ups and then allows settlement while children may still run. The 10/10 experiment validates that the improved spawn wording influences this model; it does not establish a fail-closed lifecycle barrier.

[P1] `runningDirectIds` duplicates manager lifecycle state and misses valid transitions. A fast child can settle before the ID is added after spawn, leaving a stale running ID; a settled child restarted through `subagent_send` is never re-added. Please derive this state from the manager rather than maintaining a parallel Set.

## Spec

[P1] Restarted direct children receive neither print-host wait guidance nor barrier tracking, so their required result can still be lost when the parent exits.

[P1] The fallback calls fire-and-forget `pi.sendMessage()` from `agent_settled`. Pi print mode can finish the original prompt and dispose before that nested turn completes; current tests assert pure helpers and the experiment runs all obeyed the earlier spawn guidance, so the fallback path has no end-to-end evidence.

[P2] The barrier counter is Session-global and never resets per spawn/batch, and RPC is included without protocol-level validation. The draft research records are also indexed under “Validated investigations.”

tt-a1i (commented):
审查提交：`a10103da192b64d9d00e297a361ba9d396e4ec1f`。

需要修改后再合并。

### 具体问题

**P1 — 两次提示不能形成 print-host 生命周期屏障** · [extensions/subagents/index.ts:566](https://github.com/openpi-dev/openpi/blob/a10103da192b64d9d00e297a361ba9d396e4ec1f/extensions/subagents/index.ts#L566)

agent_settled injects at most two follow-ups then returns regardless of running children; sendMessage is fire-and-forget. Prompt adherence samples cannot prove settlement waits.

**P1 — runningDirectIds 与实际子代理状态分离** · [extensions/subagents/index.ts:1070](https://github.com/openpi-dev/openpi/blob/a10103da192b64d9d00e297a361ba9d396e4ec1f/extensions/subagents/index.ts#L1070)

Set insertion happens after spawn; onSettled can run before insertion. subagent_send restart has no corresponding insertion, so restarted children are not tracked.

### 验证范围

Full production diff and research/test changes inspected for evidence scope; same head as previous self-review and both lifecycle defects still present.

### 限制与后续

No live model call or fallback print end-to-end run. Existing 10/10 experiment is prompt-following evidence only. Research records marked draft must not be indexed as validated.

PR #522, opened 2026-09-09
fix(tui): present /btw asides as self-contained child pages

### LINKED ISSUE #460: [TUI] /btw 界面与主会话重叠，且主对话末尾与 /btw 开头两处内容同时被截断丢失
## 背景与问题

   在使用 `/btw` 进行技术问答时（特别是包含多个小问题或回答篇幅较长时），遇到较严重的界面重叠与内容截断问题：

   1. **界面上下重叠粘连**：进入 `/btw` 后没有干净独立展示，而是直接压在主会话屏幕中间，上半截屏幕依然残留着上一轮聊天的内容。
   2. **交界处两头内容被截断**：
      - 主会话上一轮回答的末尾部分被直接削掉；
      - `/btw` 回答的前半部分也不见了（例如问了两个问题，第一个问题的回答完全没看见，直接从“第二大点”开始显示）；
      - 两截被切断的残片直接在屏幕中间拼贴在一起。
   3. **退回主会话才发现内容没丢**：退出 `/btw` 回到主对话，按 `Ctrl+O` 展开详情，才发现上一轮末尾和 `/btw` 开头其实都在。模型回答是完整的，但在 `/btw` 界面中两头各被吃掉了一块。

liwen-aaa:
## Problem

Fixes #460.

Opening `/btw` on a long answer misrepresented the child transcript in three
ways that compounded into the reported symptom — a page that looked pasted into
the middle of the screen with content lost at both sides of the seam:

1. **The page opened at the end of the transcript.** `TranscriptViewport`
   follows the end by default, and the page reconciled to that on its first
   render. A `/btw` answer longer than one screen therefore started partway
   through: asked two questions, the answer to the first was never visible and
   the page began at the second heading.
2. **Hidden output above was never reported.** The scroll indicator was
   `followingEnd ? "" : "↓ N"`, so the state that hides the beginning was
   exactly the state that displayed nothing. The missing opening read as lost
   output rather than off-screen output.
3. **The mouse wheel scrolled the terminal's own history.** In regular
   (non-fullscreen) mode nothing captured wheel events, so scrolling inside the
   child page moved the host terminal's scrollback instead. That slid the
   parent conversation into view above a child page still occupying the lower
   rows — the "two cut fragments pasted together in the middle of the screen"
   the issue describes.

None of this lost data: leaving `/btw` and expanding with Ctrl+O showed the full
answer. All three defects were in the operator-facing projection.

## Value

A `/btw` aside reads as a self-contained page. The answer starts at its
beginning, hidden output is discoverable in whichever direction it exists, and
the wheel scrolls the page instead of dragging the parent conversation into it.
The fix lands in `extensions/shared/agent-session-page.ts`, so Direct Subagent
takeover and Workflow child transcript pages get the same behavior.

## Approach

Three commits, each with its own tests:

- `1e11ff5` — `TranscriptViewport` gains `linesAbove()`, symmetric to
  `linesBelow()` and clamped by the same `maxScrollTop`. The page reports both
  directions in the rule row it already renders, so the overlay keeps the exact
  terminal height it declares.
- `f88ca0f` — the first render anchors at the start when the transcript already
  overflows, and keeps following otherwise. The distinction is "output produced
  before I looked" versus "output arriving while I watch": a live child still
  scrolls streaming output into view, and reaching the end explicitly resumes
  following as before. `TranscriptViewport` stays a pure follow-to-end
  primitive; the page, which knows the lifecycle, decides the opening anchor.
- `eebcd75` — wheel handling split by TUI mode. In `regular` mode the page
  enables SGR mouse reporting (`?1000h`/`?1006h`) while focused and restores it
  on blur and dispose, parsing wheel events itself. In `fullscreen` mode Pi
  already owns mouse reporting, so the page implements `handleMouse` and takes
  normalized events from the host instead of touching terminal modes. Wheel-up
  pauses following like keyboard scrolling. `TakeoverView` and
  `WorkflowDashboard` forward focus and mouse events and dispose the page so
  reporting cannot leak past the overlay.

Two existing tests encoded the old open-at-end behavior (`doesNotMatch(/↓/)`,
and `match(/line 39/)` + `doesNotMatch(/line 0/)`). They were updated to the new
contract while keeping what they were really guarding: a paused reading anchor
is not pushed by appended rows, `G` resumes following, and the indicator rides
its rule without changing overlay height.

## Validation

Windows, Node 24.16.0:

- `biome format .` — clean
- `biome lint . --error-on-warnings` — clean
- `tsc --noEmit` — clean
- `node scripts/check-config-contract.mjs` — ✓ 15 persisted fields
- `node scripts/check-discipline-ledger.mjs` — ✓ 12 append-only rows
- Vitest, all `tests/**/*.spec.ts` — **9 files, 130 tests, all pass**
- Node, all `tests/**/*.test.ts` except `background-terminals` —
  **1377 tests, 1366 pass, 2 fail, 9 skipped**
- The four suites owning this code
  (`agent-session-page`, `transcript-viewport`, `subagents/takeover`,
  `workflows/dashboard`) — **54/54 pass**

The 2 Node failures are pre-existing and unrelated, verified by reverting the
changed files to `origin/main` and re-running: identical failures.

- `file-mutation-display/index.test.ts` — "overrides all seven activity renderers"
- `git-info/process.test.ts` — "captures output and tolerates command failures"
  (`process.exitCode = 7` observed as `-1`)

New tests cover: hidden rows reported in both directions and zero when the
transcript fits; a long answer opening at its start with the question visible;
a live child still following streaming output; a busy child opening at the start
and resuming follow on `G`; regular-mode wheel capture enabled only while
focused and restored on blur/dispose (idempotent, and not re-enabled after
dispose); and fullscreen-mode `handleMouse` scrolling without writing terminal
modes.

`tests/extensions/background-terminals/manager.test.ts` was excluded because it
does not terminate on this machine, and this was diagnosed rather than assumed.
Driving `TerminalManager` directly reports:

```
Process-tree termination could not be confirmed: taskkill exited 128;
taskkill timed out after 400ms; helper closed after SIGKILL;
stdio did not close after termination
```

`taskkill /T /F` exits 128 because the `cmd.exe` wrapper has already gone while
its orphaned `node.exe` grandchild survives; after `runtime.dispose()` one
`ChildProcess` and two stdio `Socket` handles stay open, so the Node test runner
cannot exit. Reverting this branch's files reproduces the same 7 failures and
the same hang, so it is pre-existing and independent of this change. It is worth
a separate issue, since CI runs this suite on `windows-latest`.

Manual confirmation of the reported scenario: `/btw` with a two-question answer
now opens showing the question and the first heading, with `↓ N` for the
remainder; wheel scrolling stays inside the page and no longer reveals the
parent conversation.

## Impact

- **User-visible behavior:** child session pages (`/btw`, Direct Subagent
  takeover, Workflow child transcripts) open at the start of output that
  predates them, report hidden output as `↑ N` / `↓ N`, and consume wheel
  scrolling instead of moving terminal history. Live children still follow
  streaming output.
- **Model-visible context/tools:** none.
- **Runtime/lifecycle:** in regular TUI mode a focused child page toggles SGR
  mouse reporting and restores it on blur and dispose; `TakeoverView` and
  `WorkflowDashboard` dispose the page on close so the mode cannot leak.
- **Persisted config/data:** none.
- **Compatibility/risk:** low-to-moderate. Terminal mouse modes are host state,
  so it is scoped to `mode === "regular"`, made idempotent, and restored on
  every exit path, with tests pinning both modes. Overlay height is unchanged
  and asserted.

### DIFF
--- a/extensions/shared/agent-session-page.ts
+++ b/extensions/shared/agent-session-page.ts
@@ -2,7 +2,12 @@ import type {
   KeybindingsManager,
   Theme,
 } from "@earendil-works/pi-coding-agent";
-import type { Component, Focusable, TUI } from "@earendil-works/pi-tui";
+import type {
+  Component,
+  Focusable,
+  TUI,
+  TuiMouseEvent,
+} from "@earendil-works/pi-tui";
 import {
   Key,
   matchesKey,
@@ -88,6 +93,8 @@ export class AgentSessionPage implements Component, Focusable {
    * "output produced before I looked" from "output arriving while I watch".
    */
   private anchored = false;
+  private mouseReporting = false;
+  private disposed = false;
 
   private _focused = false;
   get focused() {
@@ -95,6 +102,16 @@ export class AgentSessionPage implements Component, Focusable {
   }
   set focused(value: boolean) {
     this._focused = value;
+    // In regular mode the terminal otherwise scrolls its own history, exposing
+    // the parent above a partially visible child page. Fullscreen Pi already
+    // owns mouse reporting and dispatches normalized events to handleMouse.
+    const capture = value && !this.disposed && this.tui.mode === "regular";
+    if (capture !== this.mouseReporting) {
+      this.mouseReporting = capture;
+      this.tui.terminal.write(
+        capture ? "\x1b[?1000h\x1b[?1006h" : "\x1b[?1000l\x1b[?1006l",
+      );
+    }
   }
 
   constructor(
@@ -112,6 +129,14 @@ export class AgentSessionPage implements Component, Focusable {
   }
 
   handleInput(data: string) {
+    const mouse = /^\x1b\[<(\d+);\d+;\d+([Mm])$/.exec(data);
+    if (mouse) {
+      const button = Number(mouse[1]);
+      if (mouse[2] === "M" && (button & 64) !== 0 && (button & 3) < 2) {
+        this.scroll((button & 1) === 0 ? -SCROLL_STEP : SCROLL_STEP);
+      }
+      return;
+    }
     const state = this.source.getState();
     if (this.keybindings.matches(data, "app.tools.expand")) {
       this.toolsExpanded = !this.toolsExpanded;
@@ -188,6 +213,22 @@ export class AgentSessionPage implements Component, Focusable {
       this.tui.requestRender();
       return;
     }
+  }
+
+  private scroll(delta: number) {
+    this.viewport.scrollBy(delta, this.rowCount, this.viewportSize);
+    this.tui.requestRender();
+  }
+
+  handleMouse(event: TuiMouseEvent) {
+    if (event.type !== "wheel") return;
+    this.scroll(event.wheelDelta ?? 0);
+    return { handled: true };
+  }
+
+  dispose() {
+    this.disposed = true;
+    this.focused = false;
   }
 
   private rule(width: number, left = "", right = "") {
@@ -304,9 +345,8 @@ export class AgentSessionPage implements Component, Focusable {
       transcript.length,
       transcriptCapacity,
     );
-    // Both directions are reported. Opening a child page follows the end, which
-    // scrolls the beginning of a long answer out of view; without an "above"
-    // marker that output looks lost rather than merely off-screen.
+    // Both directions are reported so hidden output is discoverable whether
+    // the reader is at the opening, following new output, or paused between.
     const overflowNote = [
       linesAbove > 0 ? `↑ ${linesAbove}` : "",
       linesBelow > 0 ? `↓ ${linesBelow}` : "",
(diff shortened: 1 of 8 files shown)


tt-a1i (approved):
Reviewed at exact head eebcd75addffd6ca4291d9d8651cdeeeda78cd2d. Spec review passes: the child page opens at the beginning of pre-existing overflow, reports hidden rows in both directions, and keeps wheel scrolling inside the page across regular and fullscreen TUI paths. The exact-head focused suites pass 54/54 and `bun run check` passes against the locked Pi 0.85.1 API. Hosted CI is green. Non-blocking architectural note: regular-mode SGR mouse capture writes terminal-global modes directly; the current focus/blur/dispose coverage and Pi overlay mode-switch guard bound the risk, but a future Pi-owned disposable/ref-counted mouse-capture seam would compose more cleanly.

PR #523, opened 2026-09-09
fix(subagents): align investigator role wording

### LINKED ISSUE #514: bug(subagents): explorer/reviewer described as read-only but inherit write/bash [under investigation]
## Status

Under investigation. Isolated reliability work against v0.8.1 (`a3edee28`). Not a formal Benchmark.

## Impact

Built-in `explorer` / `reviewer` descriptions say read-only, but they omit `tools`, so `effectiveChildToolAllowlist(agentType?.tools)` is undefined and the child inherits the parent surface (`read, bash, edit, write, …`). Spawn result then advertises those tools. A parent that requested a read-only investigator can still get a child that *can* write.

Task E samples did not create `CHILD_ESCAPED.txt` (no observed write escape). This is a presentation / allowlist gap, not yet a proven escape.

## Evidence

E-01 spawn result (high-fast): `Agent type "explorer" applied. It can only use: read, bash, edit, write, fd, rg, git_show, git_diff, git_log.`

`READ_ONLY_AGENT_TOOLS` exists in `agent-types.ts` but is not applied to the built-in explorer/reviewer roles.

## Unblock

Confirm whether inheritance-without-tools is intentional. If not, attach `READ_ONLY_AGENT_TOOLS` to explorer/reviewer (and any planning-compatible read-only role) and add a regression that the spawn result cannot list `write`/`edit`/`bash` for those roles. Then revalidate task E (forbidden file hash + no CHILD_ESCAPED) without loosening permissions.

bkyssn:
## Problem

Built-in `explorer`, `reviewer`, and `advisor` inherit the active parent child-tool surface in ordinary mode, but their role descriptions and child system-prompt addenda described them as read-only. This made the model-visible role purpose conflict with the `[inherited-tools]` capability class and the effective tools reported after spawn. Fixes #514.

## Value

Parents can select investigator roles from wording that matches the capability boundary actually enforced by Pi and OpenPI. Explicit role allowlists and Plan Mode remain the places that create a read-only child boundary.

## Approach

- Remove read-only authority claims from the three built-in investigator role descriptions and system-prompt addenda.
- Align the Subagents role reference with the existing inherited-tool contract.
- Add a focused prompt contract that exercises `capabilityClass()` through the public roster, derives effective child tools with the existing helpers, and confirms spawn reporting excludes parent-only tools. It also keeps `READ_ONLY_AGENT_TOOLS` classified as `read-only` for explicitly bounded roles.

## Validation

- `node --test --experimental-strip-types tests/extensions/subagents/prompt.test.ts` — 14 passed.
- `node --test --experimental-strip-types tests/extensions/subagents/agent-types.test.ts` — 22 passed.
- `node --test --experimental-strip-types tests/extensions/shared/child-session.test.ts tests/extensions/workflows/execute.e2e.test.ts` — 42 passed.
- `bun run check` — passed.
- `bun run test` — failed in two pre-existing, independently reproducible Plan Mode renderer cases: `tests/extensions/plan-mode/result-rendering.test.ts` leaves OSC 8 terminal-link escapes in rendered text. This diff does not touch Plan Mode or rendering code.
- Manual Pi TUI smoke — not run; the focused tests cover the model-facing role roster and spawn-result contract deterministically.

## Impact

- User-visible behavior: investigator role wording and the Subagents reference now match ordinary-mode inherited tools.
- Model-visible context: the three child system-prompt addenda no longer claim a read-only capability they do not enforce.
- Runtime/lifecycle and persisted config/data: none.
- Compatibility: built-in `implementer`, custom roles that omit `tools`, explicit allowlists, parent-only exclusions, and Plan Mode restrictions retain their existing behavior.


tt-a1i (approved):
Reviewed at exact head cd359f23f1bf6fa5cb74ce86dc2e2cb38cfa8b3d. Standards and Spec both pass with no actionable findings. The revised investigator wording now matches the runtime-enforced inherited-tool contract, while explicit allowlists and Plan Mode remain the real read-only boundaries. Focused contract tests and hosted CI are green.

PR #375, opened 2026-09-04
feat(web): expose Pi thinking state

seekskyworld:
## Problem

Adds the reasoning-state inspection slice of #348. The Web workbench could not show the active Pi thinking level or model-supported levels.

## Value

Users can inspect reasoning configuration from the browser while Pi remains authoritative.

## Approach

Expose optional runtime `getThinkingState()` backed by Pi AgentSession and an authenticated `GET /api/thinking` endpoint. Existing controllers remain compatible; no credentials or configuration writes are introduced.

## Validation

- `npx tsc --noEmit`
- `git diff --check`

## Impact

- User-visible behavior: adds bounded reasoning-state inspection.
- Model-visible context/tools: none.
- Runtime/lifecycle: read-only Pi state projection.
- Persisted config/data: none.
- Compatibility/risk: additive optional seam; setting changes remain future work.

Issue #195, opened 2026-08-25
docs(routing): define provider/model selection, fallback, reasoning state, and invocation evidence


tt-a1i:
# docs(routing): define provider/model selection, fallback, reasoning state, and invocation evidence

## Executive conclusion

OpenPI does not need a provider stack, model registry, fallback chain, reasoning store, cache identity, or endpoint router. Pi already owns those native seams.

OpenPI needs one documentation contract that distinguishes:

```text
configured preference
  ≠ Session-selected model
  ≠ resolved request route
  ≠ transport attempt
  ≠ provider-returned model/response
  ≠ successful invocation
```

and:

```text
reasoning effort
  ≠ reasoning-token usage
  ≠ visible reasoning summary
  ≠ opaque/encrypted reasoning item
  ≠ provider-native continuation state
```

Five independent source workstreams completed without failure, followed by synthesis and separate verification.

Fixed baselines:

- OpenPI `2a69d3f32994da4123f1312b7fa84ef3d6119be1`
- installed Pi `@earendil-works/pi-coding-agent@0.84.3`
- Codex `a9e447a69dee4f2789dd8d8c776e314772c1f049`
- Hermes `cbd8de8ad64530be01efea23b7764d5c37c634ed`
- Grok Build `c2ad97f87aea4303b6000a2c22128bc91ee76c9b`
- Gemini CLI `812f7a2bcf20b6e80e2e50c3c8fa8e26567bc1e8`
- OpenCode `8615731d46153dd29b89e205fb55b2cc16205cb0`
- Claude Code: official rolling documentation only; implementation internals are proprietary.

## Runtime provenance

This is a fixed-source architecture audit, not a diagnosis of the currently loaded OpenPI runtime.

Verified while preparing this Issue:

```text
source checkout: /Users/tushaokun/work/my-pi-setup
checkout HEAD:    2a69d3f32994da4123f1312b7fa84ef3d6119be1
checkout branch:  main
pi list source:   /Users/tushaokun/work/openpi-main-runtime
OpenPI sources:   exactly one
```

The inspected checkout and loaded runtime are different paths. Source conclusions below must not be represented as observed endpoint, authentication, fallback, usage or UI behavior in the running Session.

Evidence contract: [README — 开发运行时：区分 npm 与当前源码](https://github.com/tt-a1i/openpi/blob/2a69d3f32994da4123f1312b7fa84ef3d6119be1/README.md#开发运行时区分-npm-与当前源码).

### Selection and routing

[...]

PR #531, opened 2026-09-10
feat(web): add a Pi-native thinking-level selector to the composer

Matt-qwq:
## Problem

The Web workbench can inspect the reasoning/thinking state but cannot change it. `InspectionPanel` renders the current level and the available set read-only, and the composer's model picker only changes the model. There is no way to raise or lower reasoning effort for the active Session without leaving the workbench. This is the unmet `change thinking/reasoning level` half of #348: #375/#444 already landed read-only `GET /api/thinking` inspection with an explicit `unknown` fallback; the setter and the UI control are still missing.

Related to #348 (advances the thinking/reasoning criterion; does not close it), #76 (Web workbench roadmap), #455 (Web UI/UX task), and #195 (Pi owns reasoning state). Continues #375 / #444 and credits `seekskyworld` for the read-only inspection slice.

## Value

- Operators can adjust reasoning effort for the current task in-place, at the same place they pick a model.
- Stays Pi-native: levels come from `session.getAvailableThinkingLevels()` and `session.supportsThinking()`; OpenPI adds no second reasoning store, no browser-side parser, and no new persisted configuration. This matches #195 (reasoning state is Pi-owned; reasoning UI is presentation only) and #76 (configuration writes stay on the controlled Pi/OpenPI path).
- Keeps the toolbar stable and discoverable: unsupported or unselected models keep a disabled placeholder with a reason instead of hiding the control.
- Session-only by design: no global default is persisted, so no `extensions/setup/`, `setup-config.ts`, `SETUP.md`, or README default changes are required, and no second config source is introduced (cf. the #352 precedent).

## Approach

- **Protocol** (`web/protocol/types.ts`): new `WebThinkingState { level, available, supported, revision }`; optional `thinking` on `WebSnapshot`. Protocol version unchanged (optional field).
- **Runtime** (`web/runtime/pi-runtime.ts`): `getThinkingState()` gains `supported`; new `setThinkingLevel(level, { expectedSessionId })` runs through `serializeControllerMutation` (synchronous Pi setter, no retained runtime reference) and enforces a non-configurable merge-to-latest bound (≤1 in flight + 1 pending; every waiter resolves to the final authoritative projection or its own 409 `SESSION_CONFLICT`).
- **Adapter** (`web/adapter/pi-adapter.ts`): fail-open `safeThinking()` — a throwing getter omits `thinking` and never breaks `/api/snapshot`; the projection is bounded (≤16 levels, each ≤500 chars).
- **Host** (`web/host/web-host.ts`): `POST /api/thinking` placed before the non-GET fallback and made lease-sensitive (400 invalid body, 409 `WORKSPACE_REQUIRED`/`SESSION_CONFLICT`, 501 unavailable, `THINKING_LEVEL_NOT_AVAILABLE` for unsupported models). `GET /api/thinking` keeps #444's shape and never-500 `unknown` fallback; both responses now share the host-bound projection helper and carry the monotonic host `revision`.
- **Store** (`web/ui/src/store/web-store.ts`): last-intent-wins single-flight flush; a monotonic `revision` gate (scoped to the Session epoch, reset on cursor reset) so a stale snapshot or reordered response can never regress the level; `thinking_level_changed` is applied as a local revision-gated patch rather than forcing a full snapshot refresh.
- **UI** (`Composer.tsx`, `Transcript.tsx`, `InspectionPanel.tsx`, `i18n.ts`, `styles.css`): icon-only picker beside the model picker with a section title, current-level check, disabled-reason tooltip, pending visual, and a mismatch warning. The level is always present in the accessible name, and the check/`data-level` are redundant signals (WCAG 1.4.1).
- **Non-goals**: per-turn historical level annotation, global default persistence, a server-side admission cap or user-visible concurrency knob, and any new `/openpi-setup` configuration contract.

## Validation

```
bun run check            # config contract, discipline ledger, UI tsc + Vite build, format, lint, tsc — all green
bun run test             # Node 1544 pass / 0 fail / 1 skip; Vitest 164 pass
bun run test:web:e2e     # 22 passed: 11 existing + 11 thinking-picker browser matrix, axe clean
bun run test:web:e2e:provider  # 1 passed: hermetic end-to-end provider proof
```

- Committed `web/dist` was rebuilt from source and `bun run check` leaves it clean (CI's stale-build guard).
- The provider e2e seeds a local fake OpenAI-compatible model (`models.json`) and asserts the real backend request body carries `reasoning_effort: "high"` after selecting `high` in the browser, and omits the field for `off`; it also asserts the picker is disabled while a turn runs and the read path still reports `high` after reload.
- The browser state matrix covers open/section/check, zero-write on the confirmed level, one-write on change, keyboard open/arrow/enter/escape, last-intent-wins while a write is in flight, failed write → notice + `GET /api/thinking` reconcile, 501 → notice, out-of-set mismatch warning, running lock, narrow-viewport overflow, and axe.
- Real-browser manual evidence captured on a live workbench with real models (DeepSeek V4.1 Flash and GLM 5.3 Flash): placement beside the model picker, the per-level icon colors (`off/low/medium/high/max`), and the opened menus for both models (GLM exposes `medium`, DeepSeek exposes `off`).
- Not run/claimed: installed-package (`npm pack`) acceptance and the native TUI. No merge-tree check against a newer `main` was needed because the branch is based on `main@5bf2fe2`; it will be rebased if `main` moves.

### Real-browser evidence (live workbench, real models)

Captured from a running workbench with real models; each dropdown was captured after its animation settled. The picker's Brain icon is tinted by the active level. `medium` is offered only by GLM 5.3 Flash and `off` only by DeepSeek V4.1 Flash — each model exposes exactly the levels its `thinkingLevelMap` allows, so the two menus differ by design.

**Placement** — right of the model picker, left of send:

Toolbar placement

**Icon color by level** (`off` from DeepSeek; `low/medium/high/max` from GLM):

off — grey
low — blue
medium — green/teal
high — orange
max — red

**Opened menus:**

GLM menu — low/medium/high/max, medium present
DeepSeek menu — off/low/high/max

## Impact

- **User-visible behavior**: adds a thinking-level picker to the composer; disabled placeholder with a reason for unsupported/unselected models; disabled while a turn runs.
- **Model-visible context/tools**: changes the reasoning effort carried on the next provider request (proven by the provider e2e); no new model-facing tools or schemas.
- **Runtime/lifecycle**: adds a synchronous thinking setter and one lease-sensitive POST route; no retained runtime references, no persistence.
- **Persisted config/data**: None — session-only, no new setup field or preference source.
- **Compatibility/risk**: protocol adds an optional field without bumping the version; stale snapshots are rejected by the monotonic `revision`; there is intentionally no server-side rate limit (client is single-flight), documented as a known gap.


tt-a1i (dismissed):
Reviewed 351769c05c58bef3046203b2caf6fbe6c394b297 after current-main integration. Standards: no remaining actionable blocker identified. Spec: Pi owns supported thinking levels and serialized Session-bound mutation; the composer projects the resulting canonical state. No P0/P1 found. This supersedes findings against older implementations where stated; merge remains conditional on current required CI and branch protection.

tt-a1i (dismissed):
已解决最新 main 集成冲突并重新生成 Web 产物。双线复审确认保留移动侧栏焦点行为和本 PR 功能；check通过，完整测试及Web E2E继续执行。未发现P0/P1。此账户推送了集成修复，仍需另一维护者满足最后推送后的独立批准。

tt-a1i (dismissed):
追加跨环境测试修复：基础页面不再假定真实 runtime 一定不支持思考；将禁用状态及标签断言移至明确 supported:false 的夹具。原行为覆盖保留，bun run check 与27项浏览器E2E通过。最新完整CI正在验证。此账户推送了修复，仍需独立批准。

PR #332, opened 2026-09-01
fix(cron): bound schedules and due batches

6iKUN6:
## Problem

Closes #323.

`/cron` validated the numeric duration before unit conversion, so a finite input could overflow to `Infinity` in milliseconds and be reported as scheduled even though it could never become due. The Session job list and the aggregated due-job follow-up also had no total count or UTF-8 byte bounds.

## Value

Cron scheduling now fails explicitly for unrepresentable times and keeps both Session memory and model-visible due delivery within fixed limits. Valid pending work is retained instead of being silently lost when one delivery batch reaches its budget.

## Approach

- Require converted durations and initial absolute due times to be safe integers.
- Limit each Session to 64 active Cron jobs.
- Deliver at most 16 due jobs and 48 KiB of UTF-8 content per idle tick.
- Leave jobs outside the selected batch pending for later idle ticks.
- Advance only the jobs included in a successfully queued batch, preserving existing retry, one-shot, recurring, and idle-only behavior.

The limits are fixed runtime invariants rather than new user configuration. Batch selection preserves existing job order.

## Validation

- `bun run check` — passed: config contract, discipline ledger, formatting, lint, and typecheck.
- `bun run test` — passed: 1113 tests, 0 failed, 1 skipped; the additional Vitest suite passed 30 tests.
- `git diff --check` — passed.
- Manual Pi smoke after `/reload`, with `pi list` showing only the local checkout:
  - `/cron in 30s 回复：cron one-shot 已触发` scheduled and fired successfully.
  - `/cron in 9007199254741s 不应创建` was rejected with an explicit duration warning.
- Job-count, due-batch count/byte bounds, pending retention, unsafe absolute due time, and delivery retry are covered by automated tests.

## Impact

- User-visible behavior: oversized durations, unsafe absolute due times, and a 65th active job are rejected with warnings. Large simultaneous due sets are delivered across later idle polling ticks.
- Model-visible context: one Cron follow-up is bounded to 16 jobs and 48 KiB of UTF-8 content.
- Runtime/lifecycle: a Session retains at most 64 active jobs; only successfully delivered jobs advance.
- Persisted config/data: none. Cron jobs remain Session-local and in memory.
- Compatibility/risk: ordinary `in` and `every` schedules are unchanged. Only previously unbounded or unrepresentable inputs are rejected or split.

PR #538, opened 2026-09-13
fix(cron): prevent pending jobs from starving behind recurring runs

627150795:
## Problem

The bounded delivery introduced in #332 can starve accepted Cron jobs. Schedule 16 `every 30s` jobs followed by an `in 30s` job. If each eligible idle poll occurs after the recurring jobs become due again, insertion-order batching keeps selecting the first 16 and never delivers the one-shot. The UTF-8 byte limit can cause the same starvation with fewer selected jobs.

## Value

Already-due reminders make progress even when recurring jobs fill every delivery batch.

## Approach

Sort due jobs by their next scheduled run before applying the existing count and byte limits. Waiting jobs retain their older due time, while successfully delivered recurring jobs advance. Equal due times retain insertion order; delivery retry, idle-only admission, and limits are unchanged.

## Validation

- Added regression coverage at both the job-count and UTF-8 byte limits, with recurring jobs becoming due again after busy gaps. Both tests failed before the fix and passed afterward.
- `node --test --experimental-strip-types tests/extensions/cron/index.test.ts tests/extensions/cron/schedule.test.ts`: 22 passed.
- The original 10-poll reproduction now delivers the previously starved one-shot.
- `bun run check` and `git diff --check`: passed.
- [GitHub CI for commit 247480a](https://github.com/openpi-dev/openpi/actions/runs/34736298289): all jobs passed, including Node 22.19.0/24/26, Web E2E, Windows process lifecycle, and Windows full-suite isolation.
- `bun run test` on Windows / Node 24.14.0: 1,498 passed, 3 failed, 10 skipped. The failures were the existing `git-info/process.test.ts` 1-second process timeout, `workflows/narrator.test.ts` log-flood timing, and the `web-host.test.ts` test process. The narrator case and all 38 Web Host tests passed on isolated rerun. The runner stops after the failed Node partition, so its later Windows serial and Vitest partitions were not reached. Full-suite success is not claimed.
- No real-provider or installed-Pi manual smoke was performed; the deterministic tests exercise the production Cron command, poll, and delivery path with an injected clock and delivery sink.

## Impact

- User-visible behavior: older due jobs are delivered before newly due recurring runs.
- Model-visible context/tools: only the order of due prompts in existing bounded batches changes.
- Runtime/lifecycle: no new timers, queues, or background owners; only successful deliveries advance jobs.
- Persisted config/data: none; jobs remain Session-local and in memory.
- Compatibility/risk: equal due times keep their prior order. Existing 64-job, 16-per-batch, and 48-KiB limits remain enforced.


tt-a1i (approved):
Reviewed at 247480ac683b0451a1fc091ac4f2ec41cf67676f.

This fixes bounded cron delivery starvation by ordering due jobs by nextRunAt before applying the existing count and byte limits. The change stays within the existing cron lifecycle, preserves recurring-job advancement semantics, and adds focused regression coverage.

Local validation: bun run check passed; 22 focused cron tests passed. Hosted CI is green. No blocking findings.

PR #545, opened 2026-09-17
fix(tui): show cleanup approval wait instead of running Bash

### LINKED ISSUE #544: TUI: show awaiting confirmation instead of Running before guarded rm
## Problem

In the interactive Pi TUI, `workspace-cleanup-guard` intercepts a direct `rm keep.txt` targeting a pre-existing workspace file and awaits a user decision. The command has **not executed**, but the collapsed Bash tool row says `Running rm keep.txt · Ns` and continues to animate while the confirmation dialog is open. That label misstates the execution phase; the visible row also repaints throughout the wait.

## Reproduction on latest main

1. Check out `f6b49ae59605b1276b8267f2886d22c03f01533c` (2026-09-15) and verify that `pi list` reports this checkout as the only OpenPI source. Reproduced with Pi 0.85.1, Node 24, macOS.
2. Create a pre-existing `keep.txt` in an isolated workspace. Start the interactive TUI with `workspace-cleanup-guard` and `file-mutation-display` loaded.
3. Have a controlled local fake Provider call `bash` with `rm keep.txt`.
4. Leave the confirmation dialog unanswered.

**Observed:** The tool row remains `Running rm keep.txt · Ns` with spinner/timer while the guard is waiting for confirmation and the file still exists. A PTY capture had 342 `Running` updates and one confirmation prompt during the observation interval. Declining preserved the file; accepting then deleted it.

**Expected:** Distinguish waiting for the user from actual command execution. The tool row should not imply `rm` is running or keep an execution timer/activity animation during the decision. After approval, it can transition to execution; refusal/cancellation remains blocked. Do not change the confirmation choices or their default behavior as part of this issue.

## Screenshot

Running rm keep.txt shown behind pending confirmation

## Scope and pointers

- [Guard confirmation](https://github.com/openpi-dev/openpi/blob/f6b49ae59605b1276b8267f2886d22c03f01533c/extensions/workspace-cleanup-guard/index.ts#L27-L39)
- [File mutation tool rendering](https://github.com/openpi-dev/openpi/blob/f6b49ae59605b1276b8267f2886d22c03f01533c/extensions/file-mutation-display/render.ts#L79-L95)

The user-reported bright/white flashing in the original terminal video has **not** been reproduced on latest main. This issue does **not** claim that the pending row causes that flash; continue original-terminal reproduction separately before changing rendering to address flashing. No product code has been changed for this report.

testikun:
## Problem

Fixes #544. Pi emits `tool_execution_start` before OpenPI’s `tool_call` cleanup guard finishes. While a guarded `rm` is awaiting confirmation, the compact Bash row incorrectly says `Running rm keep.txt` and advances a timer, although the command has not executed.

## Value

Users can distinguish a pending decision from active Bash execution. The waiting row does not keep animating or repainted every second while a person decides.

## Approach

The cleanup guard publishes the confirmation phase on Pi’s extension EventBus with Session and tool-call identity. The TUI-only file-mutation display projects it as `Awaiting approval`, clears its execution timer during the wait, and resumes `Running` after approval. Refusal/cancellation remains awaiting until the blocked result is shown. The guard decision, confirmation choices/default, actual tool definition and execution semantics are unchanged.

The source-scoped investigation is recorded in `docs/research/ISSUE_544_CLEANUP_CONFIRMATION_UI_2026-09-17.md`.

## Validation

- `bun run check`: passed (using Bun 1.3.14 via npm exec in this shell).
- `bun run test`: 1661 passed, 1 skipped, 0 failed; Vitest 220 passed.
- Focused guard, renderer, and cross-extension EventBus tests: 22 passed.
- Pi 0.85.1 interactive TUI/PTY with a single local OpenPI source at `f6b49ae`: holding confirmation for >3 seconds showed one `Awaiting approval` row and no `Running rm keep.txt` row during the wait. Esc preserved the pre-existing file; approval resumed running and deleted it.
- Ablation: removing the event-driven invalidation brought repeated `Running` rows back, so it remains necessary.

## Impact

- User-visible: the collapsed Bash tool row distinguishes the confirmation phase and no longer shows a running timer during it.
- Model-visible context/tools: none. Tool schema, prompt metadata, execution and result are preserved.
- Runtime/lifecycle: one ephemeral, Session-scoped Pi EventBus phase signal and per-call renderer invalidation; no persistence or permission changes.
- Persisted config/data: none. Compatibility risk is limited to Pi TUI projection; headless Sessions keep native tool definitions. The separately reported white `read/grep` flashing is not reproduced on current main and is not addressed by this PR.


tt-a1i (commented):
审查提交：`e548f97d90aa52840d74f90b414a7e215dc7b6b6`。

### 验证范围

Full production, test and research-record diff inspected. Frozen head 36 file-mutation-display and workspace-cleanup-guard tests passed, including real ToolExecutionComponent event invalidation and session isolation.

### 限制与后续

No independent PTY smoke in this audit. Author documents controlled macOS TUI approval/refusal evidence with provenance; does not claim Linux visual reproduction. Approval enforcement unchanged; only operator-facing confirmation phase is projected.

PR #547, opened 2026-09-17
feat(web): browse archived and terminal Session history

testikun:
## Problem

The archived sidebar only filtered the bounded snapshot (500 Sessions), so older archives could not be found or restored there. The existing read-only terminal Pi history endpoint had no Web entry point. Refs #347; deletion/fork and full transcript recovery remain separate work.

## Value

Users can page through older archives and inspect compatible terminal Pi Sessions without transferring control or copying their history into Web storage. Empty, failed, partial-scan, and end-of-history states are explicit.

## Approach

- Use the existing authenticated archived cursor/query and terminal read-only list/detail APIs; keep current Web Sessions snapshot-backed and terminal history scoped to the selected workspace.
- Project bounded text-only terminal previews in a read-only dialog with source/origin and truncation cues. Abort stale requests on view/query/workspace changes; never expose select, prompt, edit, fork, or delete for terminal records.
- Retain restore through the existing action, guard concurrent clicks, and remove restored rows without activating them. No backend or Pi Session authority changes.
- Update README with Web-vs-terminal storage and preview boundaries. Ablation removed the old snapshot-only archive grouping, unused translations, and unnecessary type assertions; the focused tests remained green.

## Validation

- `bun run check` passed (Node 24 toolchain, base `f6b49ae59605b1276b8267f2886d22c03f01533c`).
- `bun run test` passed: Node 1658 passed/1 skipped; Vitest 223 passed.
- Real Chrome: `playwright test --config tests/web/playwright.config.ts tests/web/openpi-web.e2e.ts` passed 35/35, including multi-page archive discovery, restore/failure, read-only terminal preview and 390/1280px layout.
- Isolated `pi list` reported a single OpenPI source at this checkout; the package was not installed into the user Pi agent dir.

Terminal history desktop
Terminal history mobile

## Impact

- User-visible: paginated archived view and read-only terminal history; current Web Session controls remain unchanged.
- Runtime/lifecycle: no new owner, activation, or write path; read-only requests are cancellable in the UI.
- Persisted config/data: none; archive restoration uses the existing mutation.
- Compatibility/risk: archived view now queries the canonical bounded endpoint instead of the latest snapshot. Partial scans remain explicitly partial, not a claim of exhaustive history. Refs #347 rather than closing its remaining requirements.


tt-a1i (changes requested):
审查提交：`9d95088b2958474180829b887ac97f5549fe91b3`。

需要修改后再合并。

### 具体问题

**P2 — 终端历史请求未绑定浏览器选择的工作区** · [web/ui/src/protocol/client.ts:290](https://github.com/openpi-dev/openpi/blob/9d95088b2958474180829b887ac97f5549fe91b3/web/ui/src/protocol/client.ts#L290)

SessionHistory depends on selectedWorkspace but terminalSessions sends only search query/cursor; host adapter uses runtime.cwd. Composer setWorkspace changes browser draft only. Selecting another workspace before creating a Session therefore lists old runtime workspace records; list accepts wrong-cwd rows, while detail rejects them. Added UI repro at exact head confirms /repo rows render with workspace=/other.

### 验证范围

Full production/test diff reviewed; 47 existing UI tests pass and 4-test supplemental reproduction passes, including wrong-workspace rendering. Request/host/store path traced.

### 限制与后续

No interactive browser/native-history run. E2E additions stub list/detail API; they do not cover browser draft workspace vs runtime cwd divergence.

PR #548, opened 2026-09-17
feat(web): inspect bounded Subagent owner details

testikun:
## Problem

Subagent activity in the Web workbench is a passive chip. The manager already owns exact progress and terminal outcome, but only Background Terminals have an inspectable detail path. Refs #346.

## Value

Operators can inspect a Web Session’s delegated work, including its model, source, turn/record counts, recent tools, bounded output, failure and interruption evidence, without inventing a second resource state or adding browser control authority.

## Approach

- Extend the existing Session-scoped Web observer registry with an exact-id Subagent detail projection from `manager.view.get(id)`; the existing authenticated `/api/capabilities/detail` Host route continues to enforce the active Session identity.
- Bound title/model, cwd, error, current output and the last eight tool previews at the projection boundary; report omitted bytes/tools and truncation. A restarted running child does not present a previous run’s final text as current output.
- Wire Subagent chips into the existing InspectionPanel as text-only, read-only details. Keep terminal inspection and runtime status separate; no stop/cancel/write API is introduced.
- Ablation removed an extra queued-message field; turn count, retained record count, recent tool progress and exact outcome still satisfy this slice.

## Validation

- `bun run check`: passed (Node 24, checkout base `f6b49ae59605b1276b8267f2886d22c03f01533c`).
- `bun run test`: Node 1660 passed, 1 platform skip; Vitest 222 passed.
- Focused Host/registry 47/47 and UI 51/51 passed. Real Chrome Subagent desktop/mobile test passed, including refresh, exact Session/id GET, escaped text and no writes; isolated `pi list` showed one local OpenPI source at this checkout.
- Full real-Chrome Web suite: 33/34 passed. The existing touch-viewport test intermittently closed its sidebar after a page-level Escape and timed out looking for Workspace options; that unchanged test passed when rerun alone. PR #547 fixes this test focus race, so this PR intentionally does not duplicate its patch. The new Subagent test passed in both full and focused runs.

Subagent detail desktop
Subagent detail mobile

## Impact

- User-visible: Subagent activity chips open a bounded, read-only detail panel; existing Terminal and status panels remain available.
- Runtime/lifecycle: one synchronous owner read through the existing registry; no new controller, cancellation, or persisted state.
- Compatibility/risk: output is intentionally partial; full child transcript stays in Pi. This branch shares generated Web assets and UI files with PR #547, so it should be rebased after that PR lands. Refs #346, not a closure of Workflow/control work.


tt-a1i (commented):
审查提交：`713b058e678ed281f5cb1a3c8fd2c805b27c58e8`。

### 验证范围

Full production and test diff inspected; 47 Node observer/host tests and 51 UI tests pass. Exact resource id and Session-scoped registry are checked; output is byte-bounded, plain text, read-only, and manually refreshed. Running restarted children do not expose old final output as current.

### 限制与后续

No real child-to-browser end-to-end execution. Added Playwright test supplies synthetic snapshot/detail responses and cannot establish native owner integration.

PR #550, opened 2026-09-17
feat(web): richer model picker with always-on search and capability metadata

tt-a1i:
## Problem

The Web workbench's composer model picker only offered a search box when the snapshot catalog was truncated, and each row was a single bare identity line. With many configured models there was no way to filter the loaded list, narrow by provider, or tell models apart by capability (context window, reasoning, image input) — unlike denser pickers such as Kimi Code's model dialog.

## Value

Faster, more confident model selection in the browser: instant client-side filtering for the common case, provider chips for quick scoping, and per-model capability metadata so same-named or similar models are distinguishable without memorizing IDs.

## Approach

- `ModelPicker` keeps the existing ComplexSelector shell, server-side search (debounced, byte/count-bounded) for truncated catalogs, and option-level keyboard navigation; the search input is now always rendered and filters the loaded snapshot client-side when the catalog is complete (no server round-trip).
- Provider filter chips appear when more than one provider is present; chip state resets with the search on close.
- `WebModelSummary` gains optional `contextWindow` / `reasoning` / `imageInput`, projected by a shared `projectModelSummary` in `pi-runtime` (fields omitted when the source model lacks them, keeping the projection byte-identical for partial sources); snapshot bounding and the search endpoint inherit the fields via existing spreads.
- Option rows keep the full `modelIdentity` label as the accessible name; the metadata line is `aria-hidden` so existing name-based queries and screen-reader output stay unchanged.
- A pinned footer shows ↑↓/Enter/Esc hints; the options list is now the only scrolling region so search, chips, and hints stay visible. ArrowDown from the search input focuses the first option.
- New i18n keys (en/zh): `allProviders`, `modelMetaReasoning`, `modelMetaImageInput`, `modelPickerHints`.

## Validation

- `bun run check` — green (config contract, discipline, web build, biome format+lint, tsc).
- `bun run test` — green; includes new `tests/web/model-picker-utils.test.ts` (filter/chips/ctx formatting/meta parts), new `tests/web/model-picker.spec.ts` (client filter without server calls, chips, metadata, selection, input→option focus), and a `pi-runtime` test for metadata projection.
- `bun run test:web:e2e` — 34/34, including the updated keyboard flow (search input focused on open, ArrowDown into options).
- Manual smoke on a local workbench with a stubbed 6-model snapshot: screenshots verified search filtering, chips, metadata lines, and the pinned hints footer in the zh locale.

## Impact

- User-visible behavior: model picker gains always-on search, provider chips, capability metadata, and a hints footer; opening the picker now focuses the search input instead of the first option.
- Runtime/lifecycle: none; server-side model search path unchanged.
- Compatibility/risk: `WebModelSummary` gains optional fields only; older runtimes simply omit them and the picker renders identity-only rows. `web/dist` rebuilt and included per repo convention.


tt-a1i (commented):
审查提交：`1f7567c4502ae239282f1c20e1cbe62610afee0e`。

### 验证范围

Full source and tests diff reviewed. Frozen head: 46 Node tests and 5 ModelPicker component tests passed, covering local query, provider filter, capability metadata, keyboard focus and selection.

### 限制与后续

No live browser visual acceptance. Provider chips derive from the bounded snapshot; server query is globally limited before client provider filtering, so chips are not exhaustive catalog navigation. Empty filtered server results lack a dedicated empty-state hint. Nonblocking UX limitations; needs rebase against current main.

PR #568, opened 2026-09-19
fix(goal): preserve literal objective text in prompts

### LINKED ISSUE #567: bug(goal): prompt interpolation rewrites literal objective text
## Problem

Goal objective text is stored correctly but changes when rendered into continuation, objective-update, or budget-limit messages.

On main `f6b49ae59605b1276b8267f2886d22c03f01533c`, `renderPrompt()` uses the escaped objective as a JavaScript `replaceAll` replacement string. Valid objective text containing `$&`, `$\`` or `$'` is interpreted as replacement syntax; subsequent passes also replace literal strings such as `{{ tokens_used }}` inside the objective.

For example, a goal to fix code containing a literal `$'` can receive a copy of the prompt suffix in place of that text. A task about a template containing `{{ tokens_used }}` instead receives the current numeric usage. This is data corruption, not an intended objective update.

## Reproduction and expected behavior

A regression restores a valid stored Goal through `GoalController.restore()`, calls `kickoff()`, and inspects the actual `pi.sendMessage()` payload. The stored objective stays unchanged, but the outgoing objective differs. Three additional cases exercise all existing prompt projections.

Expected: one substitution pass over the static template. User objective content stays literal except for the existing XML escaping. No change to lifecycle, budget accounting, or prompt guidance is needed.

Unchanged production code: 23 tests pass, 4 fail. A minimal callback-based substitution fix makes all 54 Goal tests pass, including the existing native Pi fixtures.

ooiuuii:
## Problem

Fixes #567.

Goal prompts reinterpret valid objective text as JavaScript replacement syntax and then substitute template-like strings inside the objective. A goal about `$&`, `$\``, `$'`, `$$`, or `{{ tokens_used }}` can therefore reach the model with different content from the stored goal.

## Value

Keep the user's objective faithful across continuations, objective updates, and budget-limit messages, including tasks about code or templates.

## Approach

Replace the chained replacement-string passes with one callback substitution over the static template. Returned values are literal and are not rescanned. Keep the existing XML escaping and budget formatting unchanged.

The fix remains in the shared prompt renderer; it does not add a template system or change Goal policy.

## Validation

- Real owner regression: restore a valid persisted Goal through `GoalController.restore()`, call `kickoff()`, inspect `pi.sendMessage()`. Also cover all three prompt projections.
- Unchanged code: **23 passed / 4 failed**. Fixed complete Goal directory: **54 passed / 0 failed**, including existing native Pi fixture tests.
- `node --experimental-strip-types --test tests/extensions/goal/*.test.ts` (expanded file list on Windows) — pass.
- `bun run check` and `git diff --check` — pass.
- One scoped read-only code review — no actionable findings.
- Full local suite is not claimed green: this Windows host has separately reproduced baseline short-timeout failures and a background-terminal process-cleanup hang. The hanging group was not repeated for this prompt fix; GitHub CI will provide a fresh full run for this branch.

## Impact

- User/model-visible: literal objective text is preserved; existing XML escaping remains in place.
- Runtime/lifecycle: no change to Goal restoration, continuation, completion, or usage accounting.
- Tools/config/persisted data: none.
- Compatibility/risk: one existing renderer and regression tests only. Actual template budget fields still substitute; objective text that resembles a field no longer does.


tt-a1i (approved):
Reviewed at head 70676e14059284e6b4d304c92ab38ce657bdfc6d. The single-pass callback substitution preserves literal objective text without changing Goal lifecycle or accounting; focused tests and required CI are green.

PR #584, opened 2026-09-19
fix(web): bind copied session selection to file identity

### LINKED ISSUE #583: bug(web): copied session IDs alias the active session file
## Problem

Several Web session paths identify the live SessionManager by session ID alone, although copying a Pi JSONL file preserves its ID. Two distinct files can therefore be treated as the same active session.

At main `f6b49ae59605b1276b8267f2886d22c03f01533c`:

- Reading a selected copy can return the active original's transcript.
- Renaming the copy can write the original instead.
- Selecting the copy can skip the runtime switch.
- When a newer copy precedes the original, the UI's ID-only current lookup can change the requested path on the next refresh.

The repository already tests duplicate IDs as distinct archive entries. The missing boundary is the active runtime/file identity, not a proposal to rewrite session IDs.

## Reproduction

Create and persist a session through native `SessionManager`, copy its actual JSONL file into the same session directory, open the copy and append different content. The adapter's selected-file read and rename tests fail on the baseline. An authenticated loopback `/api/sessions/select` test proves the runtime switch is wrongly skipped.

Separate actual store-action regressions show the same-ID copy being selected across two refreshes and an invalid copied transcript passing the canonical-retry gate.

All fixtures use isolated temporary files and synthetic session content, not user data.

## Expected

Use ID plus the actual session file when reusing the live manager or short-circuiting selection. Preserve the existing synthetic path for an in-memory current session. The UI should consume the already-provided controller/path identity rather than derive ownership from ID equality.

No new protocol fields, identity storage, ID rewriting or permissions are required. Backend/HTTP regressions fail 3/3 before the correction; store regressions fail 2/2 before the corresponding consumer fix.

ooiuuii:
## Problem

A copied Pi JSONL session retains its embedded Session ID. Using that ID alone can read, rename, or mutate the wrong file. A UI-only selection guard is insufficient: a prompt/model/thinking request can arrive or execute after another tab activates a same-ID copy.

Fixes #583.

## Value

Confirmed selection and subsequent mutations stay bound to the actual Session file. Stale requests fail closed instead of writing to a copy, and an uncertain command receipt cannot be rebound across files.

## Approach

- Reuse a small Session ID + normalized file-path comparison for the adapter and runtime. Keep SessionManager as the persistence authority; support the existing `current:<id>` identity for in-memory sessions.
- Carry the confirmed `sessionPath` on prompt, model, and thinking HTTP mutations. The host validates it, and runtime admission/serialized execution validates again after queue waits.
- Bind prompt replay records to that path as well as command/content/image identity. Retain the existing receipt for a legitimate retry without redispatching it.
- For coalesced thinking writes, apply only the last still-valid caller's level; reject stale callers individually.
- Preserve already-admitted retained-runtime work. This is not a new cancellation policy, controller, persistence format, or cross-session lock.
- Keep copied-session reads/renames/selection path-aware; use confirmed canonical snapshots before enabling writes. Update dependent UI and browser fixtures rather than weakening the guard.

## Validation

- Native HTTP regressions: partially send a real request body, select a copied JSONL via HTTP, then finish the old request. Prompt/model/thinking reject with 409 and no runtime mutation; the same copy's valid request succeeds. Missing/invalid file identity and cross-file receipt replay are also covered. **7 tests RED before the host fix, GREEN afterward.**
- Runtime queued-race regressions: **4 tests RED before guards, GREEN afterward**, using native SessionManager copied-file identities, plus normalized-path/in-memory identity coverage.
- Combined adapter/runtime/host/race tests: **119/119 passed**.
- Final head: `0c035fd6287d12dcffc256ca7ad313b4aef58a30`. Web unit/spec tests: **19 files / 279 passed** with one worker; `bun run check` passes. A parallel attempt had one unchanged provider-settings timeout; no timeout/assertion was altered.
- Review closeout: five additional RED/GREEN recovery tests protect pending thinking state, revision ownership and stale reconciliation GETs across same-ID file switches. The final scoped full-branch structured review reports **no actionable findings**.
- Final native Windows browser suite: **49/50 passed**. The remaining terminal-disposal failure is independently reproduced on unmodified main; both first-prompt creation and delayed cross-tab receipt tests pass **2/2** in isolation. No full-repository-green claim is made.
- The old four Web E2E failures were traced to inconsistent canonical snapshot mocks (selected/current file missing from the controlled session list). All four pass after correcting those fixtures, without deleting tests or relaxing product checks.

Limitations: full local `bun run test` hit its predeclared 240-second bound after reporting setup integration and reused-PID lock failures, so it did not produce a complete aggregate. A native Windows full-browser run also exposed `Signals not supported on windows` in the unchanged interactive-terminal disposal path before prompt admission. Those independent surfaces are not changed here. No paid/live model request or user's installed Pi configuration was used.

## Impact

- Three local Web mutation endpoints now require `sessionPath`; refresh stale browser assets after upgrading. Internal runtime callers may still omit the optional expected identity.
- Newer intents must not inherit stale file-scoped pending state after canonical snapshot recovery.
- No change to stop/cancel, command discovery, workbar APIs, stored JSONL IDs, or unrelated lifecycle policies.
- Updated onto main `f991f9e` as one linear topic commit; the already-merged workbench changes are not part of this PR's diff. Tests and generated UI assets are included with the same identity invariant.


web/ui/src/store/web-store.ts
```diff
@@ -321,6 +321,7 @@ export function createWebStore(
         get().selectedWorkspace === target.workspacePath &&
         snapshot?.currentSessionId === target.sessionId &&
         snapshot.selectedSession?.id === target.sessionId &&
+        get().selectedPath === snapshot.selectedSession.path &&
```
tt-a1i, line 336: [P1] Enforce copied-session identity at the server boundary

This guard only prevents new UI actions after the store notices a path mismatch. `client.prompt`, `selectModel`, and `setThinkingLevel` still send only `sessionId`, and the host/runtime validate only that ID. Since the original and copied JSONL deliberately share an ID, a request already in flight from the original page can arrive after the runtime switches to the copy and still pass, mutating the wrong session file. Please include the selected session path (or another file-unique token) in these mutation requests and validate it again at admission/execution time; response-side epoch checks cannot undo the server-side effect.

tt-a1i (changes requested):
The path-aware UI checks are a useful projection fix, but the mutation boundary still identifies a session only by its embedded ID. A copied JSONL can share that ID, so an in-flight request from the old file can still be admitted against the newly selected file. Please carry the canonical session path/file identity through the mutation protocol and enforce it in the host/runtime, with a regression covering a request that races a switch between copied sessions.

tt-a1i (changes requested):
审查提交：`8bc6375ae65948ae0476c9ffb0675812137887d2`。

需要修改后再合并。

### 具体问题

**P1** · [web/ui/src/store/web-store.ts:1275](https://github.com/openpi-dev/openpi/blob/8bc6375ae65948ae0476c9ffb0675812137887d2/web/ui/src/store/web-store.ts#L1275)

Copied-session file identity is checked only in local projection. Prompt/model/thinking requests still carry only embedded sessionId, and server admission checks only that ID. Original and copied JSONL share it, so an already-in-flight original request arriving after selection of the copy is admitted against the wrong file. Carry canonical file identity or runtime generation through mutations and revalidate at admission.

### 验证范围

Reviewed complete production/test diff and current client/runtime request checks; 64 adapter/runtime tests passed, including copied-file projection and rename isolation. Existing exact-head finding remains applicable.

### 限制与后续

Did not independently run a delayed network race or browser E2E. Tests cover UI suppression after mismatch, not stale request admission. 

### 已核对的 CI 失败

[Run 35427046809](https://github.com/openpi-dev/openpi/actions/runs/35427046809)：Web E2E 四项失败：暗色预期得到 light、运行状态按钮禁用、归档历史元素缺失、轨迹按钮等待超时；不能归为已确认基础设施故障。

web/ui/src/store/web-store.ts
ooiuuii: Thanks — the ID-only mutation gap was real. Fixed in `0c035fd6287d12dcffc256ca7ad313b4aef58a30`, rebased/linearized onto `f991f9e` without adding a new runtime layer.

I chose the existing SessionManager file identity rather than a new generation counter: the adapter already has the authoritative file path, and a shared ID/path comparison can enforce the same boundary at each admission point.

- The client sends its confirmed `sessionPath` with prompt/model/thinking. The host requires and validates it; queued runtime operations validate again when actually executing. Stale file identities return **409 SESSION_CONFLICT** rather than being converted to prompt rejection or reaching the copy.
- Prompt receipt replay is bound to the same file. Coalesced thinking writes use the last still-valid caller, not a stale caller's level. Already-admitted retained-runtime work keeps its existing lifecycle.
- Canonical reconnect recovery also clears file-owned pending thinking/revision state, and ignores a reconciliation GET from the previous file. This avoids leaving Send disabled after an otherwise correct stale-response rejection.

Proof on the real boundaries:
- Partially send a native HTTP request, switch to a copied JSONL through HTTP, then finish the old body: all three mutation routes reject without side effects. Valid requests to the copy succeed; invalid/missing path and receipt rebinding are covered. **7 tests RED → GREEN.**
- Four queued-runtime copied-file tests also fail before the guards and pass after; five additional recovery regressions were verified RED → GREEN.
- Combined adapter/runtime/host/race suite **119/119**; final Web specs **279/279** (19 files, one worker); `bun run check` passes. Final scoped full-branch review has no actionable findings.

The four previous Web E2E failures were inconsistent canonical snapshot fixtures; these now pass without relaxing the product guard. Final full native Windows Chromium run: **49/50**. The remaining test hits `/api/sessions` **500: Signals not supported on windows** after interactive-terminal use, before prompt admission. I reproduced the identical failure on unmodified `f991f9e`: `web/host/interactive-terminal.ts` passes SIGKILL to Windows node-pty during `retain()` cleanup. That needs a separate terminal-owner follow-up; I did not mix it into this PR. With that unrelated prior terminal absent, both native first-prompt creation and delayed cross-tab receipt tests pass **2/2**.

The full repository test attempt still is not green: it reached a 240-second bound after setup integration / reused-PID lock failures. Setup integration passes alone; no unrelated timeout/assertion changes were made. Fresh GitHub CI is pending. No live paid-model request or installed-user runtime was used. The three local mutation endpoints require refreshed browser assets to send the path; persisted JSONL IDs are unchanged.

tt-a1i (approved):
Re-reviewed 0c035fd6287d12dcffc256ca7ad313b4aef58a30. The copied-session mutation finding is addressed: prompt/model/thinking requests carry the file identity, Host validates it, and queued runtime mutations revalidate before execution. Local verification passed 96 Node tests and 128 UI/protocol tests. Current CI, including Web E2E and Windows, is green. This supersedes my previous request for changes. I did not independently rerun browser E2E locally.

PR #594, opened 2026-09-19
fix(suggestions): redact complete quoted credential values

### LINKED ISSUE #593: bug(suggestions): quoted credential suffixes survive transcript redaction
## Summary

The opt-in next-action suggestion transcript redactor can leave the suffix of a labelled credential in model input when the credential is enclosed in quotes and contains spaces or punctuation.

The existing matcher recognizes optional quote delimiters, but still stops the value at whitespace, commas, semicolons, or `}`. Those characters are legal inside a quoted value. The complete quoted value should be redacted while adjacent ordinary text remains intact.

## Reproduction

Base: OpenPI f6b49ae (`0.8.1`), Node 22.22.3 on Windows. All values below are synthetic.

1. Pass a normal `read` tool-result SessionEntry containing `password="north south" next=visible` to `serializeRunTranscript` from `extensions/suggestions/src/transcript.ts`.
2. Observe the serialized tool-result text.
3. Repeat with `{"secret":"north,south;east}west","ordinary":"visible"}` or `token='north south;east,west}' next=visible`.

Expected for the first case: `password=[REDACTED] next=visible`.

Actual: `password=[REDACTED] south" next=visible`.

The same truncation occurs with escaped quote characters and backslashes. Five focused serializer regressions fail on the base revision.

## Impact and scope

When suggestions are explicitly enabled, the serialized run is passed to the configured suggestion model. This can retain fragments that the existing labelled-credential redaction was intended to remove. No real credentials, provider calls, or actual disclosure were used or observed in this reproduction.

The proposed fix is limited to correctly consuming complete quoted values in the existing best-effort redactor. It does not add new secret-detection policy, guarantee removal of arbitrary unlabelled secrets, change suggestion configuration, or rewrite Session history.

ooiuuii:
## Problem

The opt-in suggestion transcript redactor can retain a credential suffix when a labelled value contains quoted whitespace or punctuation. For example, `password="north south") next=visible` used to retain `south` in context sent to the suggestion model.

Fixes #593.

## Value

Standalone quoted labelled credentials are masked as complete values before next-action prediction. Ordinary adjacent text stays available, without claiming complete secret detection or general shell parsing.

## Approach

- Accept complete quoted values, including escaped quotes/backslashes, at existing token boundaries and closing `)`, `]`, `.` punctuation. Closing punctuation sequences must themselves end at a token boundary.
- Keep punctuation-joined shell segments such as `password="north"."south"` on the whole-token fallback; do not expose previously masked suffixes.
- Preserve recognized labels, other redaction passes, transcript budgets, and lifecycle.
- Exercise the actual serializer with tool-result and native Bash SessionEntries, both quote styles, adjacent headers, unquoted credentials, and unlabelled prose.

## Validation

Latest head: `2c3c3ab64165d4b89e82f23ba630dc9299f65fb5`.

- Original quoted-value regressions fail before the fix. Maintainer follow-up added 12 punctuation regressions that failed on the prior matcher and now pass.
- Scoped review caught a newly introduced punctuation-joined segment regression; four additional RED cases led to the token-end constraint. The final structured review reports no actionable findings.
- Node 22.22.3 suggestions tests: **71/71**; Bun 1.3.14 suggestions tests: **71/71**.
- `bun run check` and `git diff --check`: passed.
- Full local repository attempt: **1,580 passed / 5 failed / 10 skipped** in the first Node group. Four unchanged Web HTTP tests passed in isolation; the unchanged reused-PID setup-config lock test still timed out. The runner did not reach serial background-terminal tests or Vitest. This is not a full-suite pass and no unrelated assertions or timeouts were changed.
- Fresh GitHub CI is tracked separately; no claim of all-green CI until it completes.

All credentials in tests are synthetic. No live suggestion-model request was made.

## Impact

- User-visible commands, settings, tools, runtime ownership, and persisted Session history: unchanged.
- Model context: complete quoted labelled values are masked at the covered boundaries.
- The review follow-up changes only the existing matching rule/comment and serializer regressions (two files, +20/-2 from the reviewed head).
- Known boundary: ambiguous nested shell wrappers with whitespace may retain the same partial value as main. This is best-effort redaction, not an arbitrary shell parser; a compatibility regression ensures a later credential header is still masked.


extensions/suggestions/src/transcript.ts
```diff
@@ -67,14 +67,16 @@ function capped(text: string, maxBytes: number, notice: string) {
...
+  // Otherwise consume a whole token without crossing into a later field.
   return text
     .replace(/\b(Bearer|Basic)\s+[A-Za-z0-9._~+/=-]+/gi, "$1 [REDACTED]")
     .replace(
       /\b(sk-[A-Za-z0-9_-]{12,}|gh[pousr]_[A-Za-z0-9_]{12,}|eyJ[A-Za-z0-9_-]{12,}\.[A-Za-z0-9_-]{6,}\.[A-Za-z0-9_-]{6,})\b/g,
       "[REDACTED]",
     )
     .replace(
-      /(["']?(?:api[_-]?key|access[_-]?key|authorization|cookie|credential|password|passwd|private[_-]?key|secret|token)["']?\s*[:=]\s*)(["']?)[^\s,;}]+\2/gi,
+      /(["']?(?:api[_-]?key|access[_-]?key|authorization|cookie|credential|password|passwd|private[_-]?key|secret|token)["']?\s*[:=]\s*)(?:(?:"(?:\\.|[^"\\])*"|'(?:\\.|[^'\\])*')(?=$|[\s,;}])|[^\s,;}]+)/gi,
```
tt-a1i: [P1] Do not leak quoted secret suffixes before punctuation

The closing-quote lookahead accepts whitespace, comma, semicolon, or `}`, but not common closers such as `)`, `]`, or `.`. For example, `password="north south") next=ok` currently becomes `password=[REDACTED] south") next=ok` (the same leak occurs before `]` and `.`), because the regex falls back to `[^\s,;}]+` and stops at the embedded space. These transcripts are sent to the suggestion model, so please recognize the remaining punctuation boundaries (or otherwise redact the complete quoted value) and add regressions for them.

tt-a1i (changes requested):
The new quoted-value branch still misses common punctuation boundaries and falls back to token redaction, which leaks part of a whitespace-containing credential into the model transcript. Please broaden the safe closing boundary and add punctuation regressions.

tt-a1i (changes requested):
审查提交：`4fd50fee11628d48e2abf93662596d791c96c5f6`。

需要修改后再合并。

### 具体问题

**P1 — 带空格的引号凭据后接右括号或句点时仍泄露尾部** · [extensions/suggestions/src/transcript.ts:79](https://github.com/openpi-dev/openpi/blob/4fd50fee11628d48e2abf93662596d791c96c5f6/extensions/suggestions/src/transcript.ts#L79)

Exact-head repro: password="north south") next=ok -> password=[REDACTED] south") next=ok; ] and . also leak. Existing review remains valid.

### 验证范围

18 transcript tests pass; 3 additional repros confirm leakage. repro.log/test.log

### 限制与后续

No live suggestion-model call.

extensions/suggestions/src/transcript.ts
ooiuuii: Fixed in `2c3c3ab64165d4b89e82f23ba630dc9299f65fb5`.

The quoted-value branch now accepts `)`, `]`, and `.` (including sequences) only when they end the token. The boundary condition deliberately keeps punctuation-joined shell segments such as `password="north"."south"` on the existing whole-token fallback, so this repair does not introduce a new tail leak.

Validation:
- The 12 new punctuation regressions failed on the prior matcher and passed after the change; both quote styles and real `toolResult` / `bashExecution` serializer paths are covered.
- Focused suggestions suite: **71/71 on Node and 71/71 on Bun**; `bun run check` passes.
- One scoped review found the adjacent-segment regression above; after the regression test and correction, the final structured review reported no actionable findings.
- Full local suite is **not green**: 1,580 passed / 5 failed / 10 skipped in the first Node group. Four unchanged Web HTTP cases passed when rerun separately; the unchanged reused-PID setup-lock test still timed out. The later serial/Vitest groups were not reached. No assertions/timeouts or unrelated modules were modified.

This revision remains a two-file serializer/test change (+20/-2 from the reviewed head). No live suggestion-model request was made and this is not a general shell parser. Fresh GitHub CI is still required before merging.

tt-a1i (approved):
Re-reviewed 2c3c3ab64165d4b89e82f23ba630dc9299f65fb5. The reported quoted-credential leak before closing parentheses, brackets and periods is fixed, with regression coverage that also preserves masking of adjacent shell segments. All 38 transcript tests passed locally and current CI is green. This supersedes my previous request for changes.

